PR #4, opened 2018-04-03
add welcome message test

vu3rdd:
Add a proper welcome message and a round-trip serialize/deserialize test.


warner:
Nice! Let's make most of the fields optional, since the server usually doesn't include a `motd` (and I think `server_tx` timestamps were a relatively late addition, so theoretically there could be an older server that doesn't include them). Is there a way to add a unit test that exercises the no-motd case?

I'll see about getting CI set up too.

warner:
but I'll land this now

PR #22, opened 2018-04-05
get started with nameplate state machine

vu3rdd:
Just started porting the _nameplate.py. Does this PR look like the way you want it to go? It is still a work in progress (I need to add more tests), but I thought of showing it to you to get a feel for the direction I am taking.

### DIFF
--- a/core/src/events.rs
+++ b/core/src/events.rs
@@ -100,7 +100,7 @@ pub enum Event {
     N_Lost,
     N_RxClaimed,
     N_RxReleased,
-    N_SetNameplate,
+    N_SetNameplate(String),
     N_Release,
     // O is for Order
     O_GotMessage,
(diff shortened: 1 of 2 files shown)


warner:
I"m so sorry, I just completely collided with your patch :(.

OTOH, I'd also broken it by rewriting the event-management enums, so it probably wouldn't have applied anyways :(.

But yes, that's the right direction. Take a look at what I wound up landing just now: I'm experimenting with switching on the state first (and the received message inside each individual do_state_123 method), and also with keeping the nameplate inside the State object rather than as a field of the Nameplate machine. That removes the need for an `Option<String>`, and it means you can't be in a state that's supposed to know the nameplate but not actually have one. There's still some boilerplate that we can probably get rid of, but so far it's not too bad.

The process() function now returns an `Events` object, which is basically a Vec of Event objects but it knows how to convert the thing you're adding into an Event, which reduces some of the boilerplate further. I also copied over a macro from the `vec![]` docs to make it easier to create these things all-at-once. I don't like the warning that I get when I actually use `vec![]` (which is a lot: there are plenty of state transformations which are entirely local to the one machine and don't emit any other events); something about that macro is incomplete.

vu3rdd:
No problem. I will study your changes.

PR #30, opened 2018-04-17
derive_phase_key: add a test and fix a bug (in that order)

vu3rdd:
The salt in the original code was an empty bytestring instead of an all zero bytestring.


warner:
BTW I filed https://github.com/RustCrypto/KDFs/issues/15 to see about changing the HKDF API to make it easier to omit the salt (using an `Option`).

Looks good, landing momentarily (with the save-a-copy change).

PR #33, opened 2018-04-19
Initial implementation of the 'send' machine

vu3rdd:
I am not sure about the way I am doing the `phase` increments during the `send` in the `boss` module. I carry the phase as part of certain states and do the increment only when we do the `send` call.

### DIFF
--- a/core/src/events.rs
+++ b/core/src/events.rs
@@ -31,7 +31,7 @@ pub enum BossEvent {
     Scared,
     Happy,
     GotVerifier(Vec<u8>), // TODO: fixed length (sha256)
-    GotMessage(String, String, Vec<u8>),
+    GotMessage(String, Vec<u8>),
 }
 
 #[derive(Debug, PartialEq)]
--- a/core/src/events.rs
+++ b/core/src/events.rs
@@ -118,21 +118,21 @@ pub enum RendezvousEvent {
 
 #[derive(PartialEq)]
 pub enum SendEvent {
-    Send(Vec<u8>),
-    GotVerifiedKey,
+    Send(String, Vec<u8>), // phase, plaintext
+    GotVerifiedKey(Vec<u8>),
 }
 use std::fmt;
 impl fmt::Debug for SendEvent {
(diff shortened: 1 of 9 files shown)


core/src/events.rs
```diff
@@ -118,21 +118,21 @@ pub enum RendezvousEvent {
...
+    GotVerifiedKey(Vec<u8>),
 }
 use std::fmt;
 impl fmt::Debug for SendEvent {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            &SendEvent::Send(ref s) => {
-                let p = String::from_utf8(s.to_vec());
+            &SendEvent::Send(ref phase, ref plaintext) => {
+                let p = String::from_utf8(phase.as_bytes().to_vec());
```
alex: You should be able to use `str::from_utf(phase.as_bytes())` and save a copy.

core/src/key.rs
```diff
@@ -125,19 +122,19 @@ impl Key {
...
-    fn sha256_digest(&self, input: &[u8]) -> Vec<u8> {
+    fn sha256_digest(input: &[u8]) -> Vec<u8> {
         let mut hasher = Sha256::default();
         hasher.input(input);
         hasher.result().to_vec()
     }
 
-    fn derive_phase_key(&self, key: &[u8], phase: String) -> Vec<u8> {
-        let side_bytes = self.side.as_bytes();
+    pub fn derive_phase_key(side: &str, key: &[u8], phase: String) -> Vec<u8> {
```
alex: `phase` really ought to be either an `&str` or an `&[u8]`. I suspect this will let you save a copy in the caller

core/src/events.rs
vu3rdd: Thanks @alex. Just pushed a couple of changes.

vu3rdd:
I pushed the `receive` machine also into this PR.

vu3rdd:
Thanks @alex.

warner:
looks good! I'm jazzed about the progress you're making!

PR #34, opened 2018-04-24
Collapse identical match branches into one

copyninja:
* Added increment_phase function to enum Boss::State to increase the
   u32 value which is used in "phase" key of the protocol.
 * Derive copy trait for Boss::State to allow copying values.


warner:
Looks good.. I added a "cargo fmt" patch to fix CI (we're checking format too). Will merge momentarily

copyninja:
Thanks. I normally format using my editor but I guess I forgot it on last one commit. Also I use rustfmt wonder what is different between them. Thanks!

PR #35, opened 2018-04-28
WIP: Parts of allocating code and wordlist

copyninja:
This branch initially started as InputHelper but later went into completing Lister Allocator and the PGPWordlist implementation. Allocator is not yet complete one missing part is choose words from wordlist. I'm just putting this PR to get some feed back on my implementation. Especially instead of panicing wrong events, I just loop back to same state etc.

Don't merge this yet I will notify once its complete.


warner:
I only got to take a brief look, but I like it. Two thoughts: let's make the wordlist one/two words per line, it looks like the entire file is just a single line right now. And let's triple-check that the RNG being used is a proper cryptographic one: it should just be pulling from /dev/urandom (or the platform equivalent); most languages I've seen use something weak by default.

copyninja:
OK Done fixed those. First I used thread_rng uses thread local storage, as per documentation it only uses /dev/urandom for reseeding once 32Kib of random data is consumed. So I switched to OsRng which uses /dev/urandom and getrandom system call on Linux and equivalent on other OS.

copyninja:
I think now its fairly complete I replaced manual setting of code with AllocateCode event and its sucessfuly getting a code generated from wordlist. I need to write a receiver to make sure Lister works or not

warner:
Oh BTW, not sure if it was clear, but to support alternate wordlists (different languages, etc), the actual wordlist is computed by the Nameplate machine (when it receives the name of the wordlist from the server), and passed into the Input machine. We only have the one wordlist so far, and the server doesn't provide any hints yet, but it might be good to make the GotWordlist event carry the list, instead of having it hardcoded into the Input machine.

warner:
also of course we should update the tests so they pass

warner:
Cool, I'll merge this now. A few notes for followup work:

* the `I_GotNamplates` event needs to include the actual nameplate list, so Input can offer it for completion.
* in `pgpwords.json`, can we sort the hashmap by index? It looks kind of random right now. Or, better yet, let's see if we can just have a list of 2-tuples (so I guess a `Vec<[String; 2]>` ?), and build the hashmap at runtime. I don't remember why I used a hex string as an index in the python form, but now it looks kind of weird to me. (maybe I just copied that structure from the warrenguy/javacsript-pgp-word-list code and it came that way).

copyninja:
@warner  I think PGP Wordlist is meant to represent the bytes as words as explained in [Wikipedia](https://en.wikipedia.org/wiki/PGP_word_list). So I think it should have bytes as key. And about sorting the word list, I actually dumped the python `raw_words` into file using json module and hence the random order. Of course we can manually create file by copying words from python.

Since it needs to represent bytes as key and word list I don't think we can represent it as vector and convert to hashmap. 
@vu3rdd  what do you think?

copyninja:
@warner I will look at I_GotNameplates and modify it as needed.

Issue #24, opened 2018-04-06
build InputHelper API


warner:
As described in #8, the tab-completion type-in-the-code API uses an object named the `InputHelper`. In the sans-io approach, this object will be defined (and owned) by the IO/API glue layer, but the Core will define some API input and output events to communicate with it.

The [Python API docs](https://github.com/warner/magic-wormhole/blob/0.10.5/docs/api.md#code-management) specify that you obtain a helper from the wormhole object with a synchronous function like`helper = w.input_code()`, and then defines the Helper's API as follows:

* `refresh_nameplates()`: requests an updated list of nameplates from the
  Rendezvous Server. These form the first portion of the wormhole code (e.g.
  "4" in "4-purple-sausages"). Note that they are unicode strings (so "4",
  not 4). The Helper will get the response in the background, and calls to
  `get_nameplate_completions()` after the response will use the new list.
  Calling this after `h.choose_nameplate` will raise
  `AlreadyChoseNameplateError`.
* `matches = h.get_nameplate_completions(prefix)`: returns (synchronously) a
  set of completions for the given nameplate prefix, along with the hyphen
  that always follows the nameplate (and separates the nameplate from the
  rest of the code). For example, if the server reports nameplates 1, 12, 13,
  24, and 170 are in use, `get_nameplate_completions("1")` will return
  `{"1-", "12-", "13-", "170-"}`. You may want to sort these before
  displaying them to the user. Raises `AlreadyChoseNameplateError` if called
  after `h.choose_nameplate`.
* `h.choose_nameplate(nameplate)`: accepts a string with the chosen
  nameplate. May only be called once, after which
  `AlreadyChoseNameplateError` is raised. (in this future, this might
  return a Deferred that fires (with None) when the nameplate's wordlist is
  known (which happens after the nameplate is claimed, requiring a roundtrip
  to the server)).
* `d = h.when_wordlist_is_available()`: return a Deferred that fires (with
  None) when the wordlist is known. This can be used to block a readline
  frontend which has just called `h.choose_nameplate()` until the resulting
  wordlist is known, which can improve the tab-completion behavior.
* `matches = h.get_word_completions(prefix)`: return (synchronously) a set of
  completions for the given words prefix. This will include a trailing hyphen
  if more words are expected. The possible completions depend upon the
  wordlist in use for the previously-claimed nameplate, so calling this
  before `choose_nameplate` will raise `MustChooseNameplateFirstError`.
[...]

PR #36, opened 2018-05-04
WIP: Input machine implementation

copyninja:
This is an attempt to port Input machine as per python magic-wormhole code. But this is not complete as it gave out several questions which needs answering.

1. We do not need another helper object as in python code, at least that is what I feel. One reason for this is unlike python code we only emit events and no IO is performed on our side. Also in python code helper is passed to rlcompleter class to trigger event that leads to return of completions. I think with some thinking we can do that in Input struct itself.
2. I've added 2 events specifically for tab completions this should be triggered only when user hits tab. But I'm unsure how we communicate back the results to completer. May be new IO Events?.

I looked at rustyline, which was suggested crate in one of the comments in #24. I see one possibility is to make Input machine implement the `Completer` trait and let it internally check for the state and depending on user input and state of the machine it returns appropriate results. This will avoid the tab completions to interact or affect the Input state machine itself.  But given that we follow sans-io approach relying on a specific I/O crate might not feel like correct. I also checked most other available crates for readline implementation and all of them needs implementation of trait like `Completer` in rustyline, only with different input and output value.

At the moment I'm bit lost. I've all what is needed to do the completion but I don't know how to really go further :-). 

@warner @vu3rdd can you please share your thoughts?.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -19,3 +19,4 @@ rustc-serialize = "0.3"
 [dev-dependencies]
 ws = "0.7"
 url = "1.7"
+rustyline = "1.0"
--- a/core/examples/readline.rs
+++ b/core/examples/readline.rs
@@ -8,8 +8,10 @@ extern crate ws;
 use magic_wormhole_core::{APIAction, APIEvent, Action, IOAction, IOEvent,
                           TimerHandle, WSHandle, WormholeCore};
 
-use std::cell::{RefCell, RefMut};
-use std::rc::Rc;
+use std::sync::{Arc, Mutex, mpsc::{channel, Sender}};
+use std::thread::{sleep, spawn};
+use std::time::Duration;
+
 use url::Url;
 use rustyline::completion::{extract_word, Completer};
 
@@ -18,12 +20,12 @@ const APPID: &'static str = "lothar.com/wormhole/text-or-file-xfer";
 
 struct Factory {
     wsh: WSHandle,
-    wcr: Rc<RefCell<WormholeCore>>,
+    wcr: Arc<Mutex<WormholeCore>>,
 }
 
 struct WSHandler {
     wsh: WSHandle,
-    wcr: Rc<RefCell<WormholeCore>>,
+    wcr: Arc<Mutex<WormholeCore>>,
     out: ws::Sender,
 }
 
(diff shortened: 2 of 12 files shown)


warner:
I'll add some thoughts to #24, I think that'll be the best place to design it. But yeah, the InputHelper will be an API-side object, communicating with our state machines through API/IO events, so it won't go into into the "core" crate.

copyninja:
@warner What do you think of my last 2 commits?. I read through your explanation on #24 but I found a bit simpler approach. Its not fully completed but I wanted to get your thoughts on this.

I've now a InputHelper object which works on InputHelperEvent (APIEvent). Main purpose of InputHelper is to stash the nameplates and wordlist. It also provides api to provide word completion and emits possible events to Input machine.  Main api which should be used by rustyline or similar library traits is `get_completions` in `Wormhole` object. This calls InputHelper's `get_completions` and processes all event before passing over result to app.

There are some pending things like what happens if the helper is called before actually machine is initialized etc but that can be fixed. And I'm also yet to write a example code on how to use this code from application side.

copyninja:
@warner I wrote down my idea of using inputhelper in readline.rs example. But at the moment it panics as it is hit by both completer and client trying to hold mutable reference, which you mentioned in the comment on #24.

copyninja:
So I handled the panic condition by using nested scopes to avoid trying to hold multiple mutable references. It looks quirky but I could only find this as the way. While trying to make that work I faced couple more issues one was the completer logic needs to handle sending out IOActions generated while trying tab completions (like TxList). This now I handled again looks to me like more of hack. 2nd was the breaking of words during hitting tab does not seem to be correct. I guess I need to use - as break character but not sure about it need to see how rustyline works. Third I need to use thread in readline examples to make sure taking input goes into separate thread as other messages does not get handled like nameplates from server etc. I'm not sure what other problems this will open, let me explore a bit more.

copyninja:
@warner Now code is able to make communication between core - input helper - app. My completion logic is flawed but still it can fetch list of nameplate from server and provide tab completion. I will look more into how to get code completion working in proper way.

warner:
Looks pretty good.. let's land this now and figure out the fixes later.

copyninja:
Thanks for merging this!. I think core part is fine just completion logic needs fixing. First I need to figure out how rustyline basically works and what BREAK_CHARS I need to use. I will look into those now and clean up warnings from code.

PR #37, opened 2018-05-12
mark a test so it is run

alex:


warner:
landing (in rebased form) shortly.. thanks!

PR #39, opened 2018-05-14
Clear all compile time warnings

copyninja:
This will help us to catch new warnings and errors in code we right easily. With
all compile warnings before useful stuff get pushed down the pile of already
existing hundreds of warnings and we miss to see warnings in our newly written code.

I've added some `#[allow(dead_code)]` in few modules, but with a **TODO** mentioning this should be dropped once the core is complete. Now since we have incomplete code these warnings will come up as most of modules are not public (`pub mod` is not used in `lib.rs`)


warner:
excellent!

PR #41, opened 2018-05-17
Use new API in readline and implement tab completion example

copyninja:
Now the readline.rs uses the new InputHelper API and successfully does the tab completion for entering code. It is also possible to receive message from python but I'm not sure if its successfully decrypted or not. 

But main idea to hook into Input machine using InputHelper api to get tab completion is now working. For this I've done couple of changes to Input machine
1. Stash the nameplate coming in from ChoseNameplate event raised from InputHelperChooseNameplate
2. input_helper_commited_nameplate method in wormhole object to access nameplate inside the Input machine which is needed by tab completion logic.

### DIFF
--- a/core/src/allocator.rs
+++ b/core/src/allocator.rs
@@ -1,5 +1,5 @@
 use events::{Events, Wordlist};
-use std::rc::Rc;
+use std::sync::Arc;
 
 // we process these
 use events::AllocatorEvent::{self, Allocate, Connected, Lost, RxAllocated};
@@ -15,8 +15,8 @@ pub struct Allocator {
 enum State {
     S0AIdleDisconnected,
     S0BIdleConnected,
-    S1AAllocatingDisconnected(Rc<Wordlist>),
-    S1BAllocatingConnected(Rc<Wordlist>),
+    S1AAllocatingDisconnected(Arc<Wordlist>),
+    S1BAllocatingConnected(Arc<Wordlist>),
     S2Done,
 }
 
@@ -78,7 +78,7 @@ impl Allocator {
     fn do_allocating_disconnected(
         &self,
         event: AllocatorEvent,
-        wordlist: Rc<Wordlist>,
+        wordlist: Arc<Wordlist>,
     ) -> (Option<State>, Events) {
         match event {
             Connected => (
@@ -92,7 +92,7 @@ impl Allocator {
     fn do_allocating_connected(
         &self,
         event: AllocatorEvent,
-        wordlist: Rc<Wordlist>,
+        wordlist: Arc<Wordlist>,
     ) -> (Option<State>, Events) {
         match event {
             Lost => (
--- a/core/src/api.rs
+++ b/core/src/api.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::error::Error;
 use std::fmt;
 
 #[derive(Debug, PartialEq)]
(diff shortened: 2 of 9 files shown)


copyninja:
@warner I merged your changes and made readline example to now work with changes to Wordlist API.

Though I had to change from `Rc` to `Arc`. We can not use `Rc` in core as I/O layer can use thread or tokio where `Arc` is unsuitable. Alternatively we should use method of using `Wordlist` as unit structure (empty structure) passed around just as place holder and where needed use the actual wordlist to get completion so we can completely keep core free of any syncrhonization stuff.

core/src/allocator.rs
```diff
@@ -15,8 +15,8 @@ pub struct Allocator {
 enum State {
     S0AIdleDisconnected,
     S0BIdleConnected,
-    S1AAllocatingDisconnected(Rc<Wordlist>),
-    S1BAllocatingConnected(Rc<Wordlist>),
+    S1AAllocatingDisconnected(Arc<Wordlist>),
+    S1BAllocatingConnected(Arc<Wordlist>),
```
warner: wait, why do we need Arc? The Wordlist should be immutable (so it should be safe to share across threads), and under what circumstances could it be shared among threads anyways? I'm willing to believe it should be Rc, but can we avoid making it Arc?

core/src/api.rs
```diff
@@ -21,6 +23,40 @@ pub enum InputHelperError {
     AlreadyChoseWords,
 }
 
+impl fmt::Display for InputHelperError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
warner: nice!

core/src/input.rs
```diff
@@ -20,14 +21,15 @@ enum State {
...
-    WantCodeHaveWordlist(String, Rc<Wordlist>), // (nameplate, wordlist)
+    WantCodeHaveWordlist(String, Arc<Wordlist>), // (nameplate, wordlist)
     Done,
 }
 
 impl Input {
     pub fn new() -> Input {
         Input {
             state: State::Idle,
+            nameplate: String::new(),
```
warner: I really think the nameplate wants to live inside the state, not next to it. Some states don't know what the nameplate is, so really that `nameplate:` ought to be an `Option` anyways. And keeping it in two separate places means they might get out of sync.

If we really need a function to say whether we've committed to the nameplate or not, it should switch on which state we're in, rather than using a separate state variable. If we're in one of the `WantCode` states or `Done`, then we're committed to a nameplate. If we're in `Idle` or one of the two `WantNameplate` states, then we're still free to choose.

But I think my intention was that the input helper keeps track of whether it's been told about a nameplate, rather than using the `Input` machine's internal state (it seems cleaner to me to keep the `Input` struct's fields private).

core/src/allocator.rs
copyninja: Rc is not thread safe!. Arc is thread safe. From the rust standard library documentation for Rc
 > A single-threaded reference-counting pointer. 'Rc' stands for 'Reference Counted'.
From  documentation of Arc

> A thread-safe reference-counting pointer. 'Arc' stands for 'Atomically Reference Counted'.

So since core can be used from either single threaded or multithreaded or tokio environment it should use Arc. I'm not happy with its use but otherwise we should use unit strcture as marker and create wordlist only when its needed (like it was before) instead of passing it around.

core/src/input.rs
copyninja: Yeah I can rework the code to see if its possible to get nameplate out of the state and drop keeping it inside the state machine itself. After writing code similar thought came to me so I will try and rework the code a bit.

copyninja:
@warner I rebased with master and fixed input.rs as per your comment. It no longer stashes nameplate internally and uses current state to get nameplate.

warner:
Ok, so the `Arc` is to protect the mutable *reference count*, even if the rest of the `Wordlist` is immutable, got it. I'm still hoping we can avoid the need for that (by having all the code that touches a `Wordlist` live in the same thread, including the completion logic), but I'm willing to believe that it can't work that way, and in any case we need to understand the IO layers better before we can figure that stuff out.

I'm still hoping we can remove the extra is-the-nameplate-committed function by having the completion code just remember whether it's called `choose_nameplate()` yet or not. But we can fix that up later. I'll add a note.

warner:
I'm merging this PR from about 15 minutes ago, before that "Use regex crate to validate integer phase" commit. Could you add that to a separate PR?

copyninja:
Ok I can do but it kinda is related. It makes demo print message from sender

warner:
merged as of 53e91d5

copyninja:
But yeah there is still pending part which I intend to add tomorrow so I'll
raise another PR for that.

PR #43, opened 2018-05-22
Io blocking

warner:
this adds a plausible IO layer that implements a blocking API. The crate and the examples compile cleanly, but don't work yet. I'm going to land this so other folks can take a look, and to keep track of API changes that might break it. (note that it doesn't do anything with InputHelper or completion yet, in fact `set_code()` is the only API that's wired up so far).


warner:
BTW this uses the [ws crate](https://ws-rs.org/guide), which uses Mio under the hood, but that's not exposed to the application. Internally, there's a separate thread for the main WormholeCore, plus one more for each timer, plus two more for each websocket (one in each direction). That's a lot of threads, but this isn't a very IO-intensive library, so I wouldn't expect to see performance problems. I expect it'd work fine for production apps. The main reason for wanting a Tokio-based API would be to integrate better with non-blocking apps, since with this blocking IO layer, such apps would need to use yet another thread to call the wormhole APIs.

PR #45, opened 2018-05-23
Introduce PeerMessage Representation and complete readline demo

copyninja:
I introduced PeerMessage representation which currently allows serializing and deserializing offer messages. With this now readline demo can display message received from python wormhole command.


warner:
looks good!

PR #51, opened 2018-06-02
Clippy fixes

copyninja:
All most all clippy warnings are handled except one triggered on use of `events![]`. This happens because when we do empty events![] it leads to below expanded code

```rust
let tmpvec = Events::new();
tmpvec
```

I tried to silence it by inserting  `#[cfg_attr(feature = "cargo-clippy"), allow(let_and_return)]` but that did not work, to silence it I guess I need to add the above line every where we use `events![]`, which I felt was over kill. So for now I've kept that open.

### DIFF
--- a/core/examples/readline.rs
+++ b/core/examples/readline.rs
@@ -11,7 +11,8 @@ use magic_wormhole_core::{APIAction, APIEvent, Action, IOAction, IOEvent,
 
 use std::error::Error;
 use std::io;
-use std::sync::{Arc, mpsc::{channel, Sender}};
+use std::sync::{mpsc::{channel, Sender},
+                Arc};
 use std::thread::{sleep, spawn};
 use std::time::Duration;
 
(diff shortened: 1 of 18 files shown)


copyninja:
Every time I foget to run `cargo test` before raising PR. *sighs*

warner:
I'm guessing that your `cargo fmt` is using a different `max_width=` (maximum line length beyond which it tries to wrap expressions) than travis. We have a `.rustfmt.toml` file that should provide that value (80 columns, I think the default is 100 which was too wide for my tastes). Do you maybe have something in your environment that's overriding that file?

I've re-run `cargo fmt`, and I also reverted one change in `blocking.io` which clippy suggested because I'd commented out the error case. Printing the error to stdout is not the right thing to do, but we need to do *something* sensible with that error case, and the `if let` form removes the reminder. For now I've uncommented the print anyways so that clippy won't think that the Err branch is unused.

I'm not sure what to do about `events![]` either. I'll ask at the next local Rust meetup.

PR #52, opened 2018-06-03
Peer messages moved to transfer module and examples adapted

copyninja:
As discussed earlier on IRC, I've moved all the peer messages into transfer module. I've created some new methods to help creating message types like one we have in server_messages module. Examples are now adapted and successfully receives or sends message from python counterpart.

Some pending parts are adding support for transit messages which I'm currently working on. This should be merged after #51 as its based on that branch.

### DIFF
--- a/core/examples/ws.rs
+++ b/core/examples/ws.rs
@@ -1,8 +1,8 @@
 extern crate magic_wormhole_core;
 extern crate url;
 extern crate ws;
-use magic_wormhole_core::{APIAction, APIEvent, Action, AnswerType, IOAction,
-                          IOEvent, OfferType, PeerMessage, WSHandle,
+use magic_wormhole_core::{message, APIAction, APIEvent, Action, AnswerType,
+                          IOAction, IOEvent, PeerMessage, WSHandle,
                           WormholeCore};
 use std::cell::RefCell;
 use std::rc::Rc;
@@ -77,9 +77,7 @@ impl ws::Handler for MyHandler {
         // TODO: this should go just after .start()
         let actions = wc.do_api(APIEvent::AllocateCode(2));
         process_actions(&self.out, actions);
-        let offer = PeerMessage::Offer(OfferType::Message(
-            "hello from rust!".to_string(),
-        )).serialize();
+        let offer = message("hello from rust!").serialize();
         // let offer = json!({"offer": {"message": "hello from rust"}});
         // then expect {"answer": {"message_ack": "ok"}}
         let actions = wc.do_api(APIEvent::Send(offer.to_string().into_bytes()));
--- a/core/examples/ws.rs
+++ b/core/examples/ws.rs
@@ -151,6 +151,9 @@ fn process_actions(out: &ws::Sender, actions: Vec<Action>) {
                                 }
                             }
                         },
+                        PeerMessage::Transit(..) => {
+                            println!("transit message recieved")
+                        }
                     }
                 }
                 _ => println!("action {:?}", api),
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -31,6 +31,7 @@ mod server_messages;
 mod terminator;
 #[cfg(test)]
 mod test;
+mod transfer;
 mod util;
 mod wordlist;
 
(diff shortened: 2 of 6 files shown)


warner:
Oh and doublecheck that your checkout includes a `.rustfmt.toml` file with `max_width=80`, so travis doesn't complain about your `cargo fmt` stuff.

copyninja:
@warner rebased and now cargo fmt worked!. Hope it won't break on travis again. .rustfmt.toml was there always, I'm not sure why it was not working yesterday. I pulled your changes from master and rebased and ran cargo fmt and now it just works.

copyninja:
Seriously?. Why is cargo fmt on my system is behaving differently than one on travis?.

warner:
maybe strace it and see if it's reading from some secondary config file?

Also, maybe we're running different versions of the tool? Look in the .travis.yml, we pin the specific version for consistency. I now forget how exactly I installed my local copy, I know there's a version that requires nightly, and there's one that works on 'stable' (which must be the one I'm using since I just run `cargo fmt` and not `cargo +nightly fmt`). Something about `rustfmt-preview`?

When I run `rustfmt --version` I get "rustfmt 0.4.1-stable (7a807262 2018-04-20)"

copyninja:
Ah we have different versions, I've `rustfmt 0.4.2-stable (febbb36 2018-04-12)`. I did not pin it rustup always updates it when I run rustup update. Should I pin it or shall we use latest ?

warner:
Oh, let's use the newest one. Or rather, let's commit a change that changes the format of the source code and simultaneously changes our .travis.yml settings to pin the new version to match. Then at home we should do whatever is necessary to agree with what travis is doing.

I run `cargo install-update -al` regularly to see what's out of date, as well as `rustup update`, but neither mentioned my rustfmt being old. I'll have to investigate how I installed that and why it's not being updated.

warner:
I've got two different versions installed here at home, and I don't know why:

* on linux (ubuntu/artful), rustfmt --version reports `rustfmt 0.4.2-stable (febbb36 2018-04-12)`
* on my Mac, it reports `rustfmt 0.4.1-stable (7a807262 2018-04-20)`
* on Travis, it reports `rustfmt 0.4.1-stable (7a80726 2018-04-20)`

In all cases, I/we installed it by using `rustfmt component add rustfmt-preview` per the instructions at https://github.com/rust-lang-nursery/rustfmt#installation .

In all cases, `rustup --version` shows 1.11.0 and `cargo --version` shows 1.26.0

My linux box (rustfmt-0.4.2) and apparently your box (0.4.2) want shorter lines. My mac (0.4.1) and travis (0.4.1) want longer lines. So at least it's consistent.

Why are my two machines giving me different versions of rustfmt? I'm starting to think that the published binaries don't match: the "rustfmt-preview-x86_64-apple-darwin" and "rustfmt-preview-x86_64-unknown-linux-gnu" components include different versions. But 1: I don't know what URL to look at to see what's being published, and 2: I don't understand why travis (an ubuntu system) is getting something different than my linux laptop (also ubuntu).

In desperation, I deleted rustup/cargo on my mac (`rustup self uninstall`) and reinstalled from scratch. Now I've got rustfmt-0.4.2 and it's wrapping things the same way as my linux laptop and like your system. So now we're all screwed and travis is the one thing behaving differently :).

copyninja:
I've same version of rustfmt across a windows machine and Linux laptop. I think we should also consider using circleci along with travis.

warner:
Ok, I changed .travis.yml to pin rust-1.26.1 instead of rust-1.26.0, and that gets the newer rustfmt-0.4.2 . I've pushed a change that reformats the code according to rustfmt-0.4.2 rules to match.

If you rebase this PR again, I think it should pass now.

I guess I just hadn't run `rustup update` on my Mac recently enough to get 1.26.1/0.4.2, and of course travis was pinned to 1.26.0 . I assumed that they wouldn't change something like rustfmt on a point release, but now that I think about it, that seems perfectly reasonable.

copyninja:
Ok now I rebased it with latest changes. Hopefully this should work fine now. `cargo fmt` locally did not show anything new.

copyninja:
OK now rustfmt passes!. So PR is now good to merge.

warner:
Great, thanks.. landed. I'm not 100% sure about having the io/blocking examples try to parse their messages as if they were file/directory/text transfers. I think that once we make a separate crate for the application (and move transfer.rs into it), I might reduce those examples back to their core of just sending/printing bytestrings, to make them better examples of the core functionality. Then the actual file-transfer apps can serve as examples of the parsers/serializers.

PR #56, opened 2018-06-12
fix warning when using empty 'events![]'

warner:
Thanks to @tarcieri for his help.


copyninja:
Great! Why it didn't come to my mind :). Now clippy should be happy with it.

PR #74, opened 2019-12-27
Avoid stringly-typed APIs in several spots

### LINKED ISSUE #49: use Display, not custom to_string, for Mood
@copyninja mentioned (in https://github.com/warner/magic-wormhole.rs/commit/46d31b82ff5faa5dd4082878ca1cdd62c520a25f#commitcomment-29139497) that `to_string` should be part of the standard `Display` trait, but I used it as a standalone method on the `Mood` struct. I'm up for making this `Display`, but the specific strings used there are part of the wire protocol, so they're fixed and should not be changed in the future (e.g. as part of i18n language-specific translations). So we need some comments to that effect, and some unit tests that ensure the strings remain fixed.

notriddle:
From the "let the knife do the work" school of Rust, this pull request has a net reduction of about twenty lines by avoiding manual stringification in favour of implementing the serde traits directly on the corresponding API types.

For more information on some of what I'm doing here, read <https://serde.rs/container-attrs.html>.

Also fixes #49; instead of implementing Display, it implements Serialize.

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -1,7 +1,5 @@
 use std::collections::VecDeque;
 
-use hex;
-
 #[macro_use]
 mod events;
 mod allocator;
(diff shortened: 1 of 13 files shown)


warner:
thanks! that's awesome, I learn more about Rust every day :)

Issue #63, opened 2018-12-11
zero keys/secrets upon Drop


warner:
I've seen curve25519-dalek use some tools/libraries to zero out the memory of sensitive data structures (keys, intermediate values used to derive keys) just before they get dropped. We should do something similar. A lot of our temporaries are on the stack, so we should zero those too.

Basically, if the application above us is careful with their own memory, then our library should not introduce any additional exposure of secrets.

I think @david415 had some ideas on this one.

david415:
seems like clear_on_drop is the popular choice:
https://crates.io/crates/clear_on_drop

I have not yet tried the "clear_stack_on_return" in my own project and so far have only used this crate to clean secrets on the heap:

#[derive(Clone, Debug, PartialEq)]
pub struct PrivateKey {
    public_key: PublicKey,
    private_bytes: ClearOnDrop<Box<[u8; KEY_SIZE]>>,
}

Whereas clear stack on return seems straight forward:

https://docs.rs/clear_on_drop/0.2.3/clear_on_drop/fn.clear_stack_on_return.html

warner:
I also see Tony's [`zeroize`](https://github.com/iqlusioninc/crates/tree/develop/zeroize), which is pure-Rust. `clear_on_drop` is used by `curve25519-dalek` (a hearty recommendation!) but we'd need to use its `no_cc` feature-flag to avoid causing problems compiling to WASM (#2).

warner:
It sounds like `curve25519-dalek` is [planning](https://github.com/dalek-cryptography/curve25519-dalek/pull/236) to move from `clear_on_drop` to `zeroize` once the latter has a stable 1.0 release, which may be soon-ish. So let's plan on using `zeroize`.

warner:
Ok, that zeroes out the main `Key` struct, and all the "phase keys" we derive from it.

Should we change the `Receive` code to return `Zeroized<Vec<u8>>` to the application? I'm thinking that if we allocate it, we should be responsible for making sure it gets erased. But I'm not sure if that would be a hassle for application authors. In the tests, I had to use a funny-looking `hex::encode(&*derived_key)` to deal with it not being a plain `Vec<u8>`. I don't know if I just couldn't find a better way, or if maybe `&*` is entirely normal and no big deal, but it looked pretty weird and redundant to me.

piegamesde:
[...]

PR #81, opened 2020-02-05
zeroize Key and derived keys upon drop

warner:
refs #63

@tarcieri hey, am I using this right?

### DIFF
--- a/src/core/receive.rs
+++ b/src/core/receive.rs
@@ -1,7 +1,6 @@
 use super::events::{Events, Key, Phase, TheirSide};
 use super::key;
 use log::trace;
-use zeroize::Zeroize;
 
 // we process these
 use super::events::ReceiveEvent;
@@ -106,9 +105,8 @@ impl ReceiveMachine {
         phase: &Phase,
         body: &[u8],
     ) -> Option<Vec<u8>> {
-        let mut data_key = key::derive_phase_key(&side, &key, &phase);
+        let data_key = key::derive_phase_key(&side, &key, &phase);
         let data = key::decrypt_data(&data_key, body);
-        data_key.zeroize();
         data
     }
 }
--- a/src/core/receive.rs
+++ b/src/core/receive.rs
@@ -106,8 +106,7 @@ impl ReceiveMachine {
         body: &[u8],
     ) -> Option<Vec<u8>> {
         let data_key = key::derive_phase_key(&side, &key, &phase);
-        let data = key::decrypt_data(&data_key, body);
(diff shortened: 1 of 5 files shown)


src/core/receive.rs
```diff
@@ -104,8 +106,10 @@ impl ReceiveMachine {
         phase: &Phase,
         body: &[u8],
     ) -> Option<Vec<u8>> {
-        let data_key = key::derive_phase_key(&side, &key, &phase);
-        key::decrypt_data(&data_key, body)
+        let mut data_key = key::derive_phase_key(&side, &key, &phase);
```
tarcieri: You can use `Zeroizing` to do zeroize-on-drop automatically: https://docs.rs/zeroize/latest/zeroize/struct.Zeroizing.html

```suggestion
        let mut data_key = Zeroizing::new(key::derive_phase_key(&side, &key, &phase));
```

src/core/send.rs
```diff
@@ -42,10 +43,11 @@ impl SendMachine {
                 match event {
                     GotVerifiedKey(ref key) => {
                         for (phase, plaintext) in self.queue.drain(..) {
-                            let data_key =
+                            let mut data_key =
```
tarcieri: Can use `Zeroizing` here (and below) as well

src/core/receive.rs
warner: Cool, could I change the signature of `derive_phase_key` to do that automatically for everything it returns? It currently returns a `Vec<u8>`.

warner: what would the new signature be? `-> Zeroizing<Vec<u8>>` ?

tarcieri: Yup

warner: ok, dumb question, what's the cleanest way to convert the unit test at line 280-ish, from `assert_eq!(hex::encode(dk11), "abcabc..")`? That `dk11` was a `Vec<u8>`, now it's a `Zeroizing<Vec<u8>>`. The compiler is complaining that `std::convert::AsRef<[u8]>` is not implemented (`hex::encode` is `pub fn encode<T: AsRef<[u8]>>(data: T) -> String`).

warner: huh, `hex::decode(&*dk11)` worked, but surely that's silly?

tarcieri: It's doing deref coercion

warner: ok, so not silly (or at least not too silly to use). thanks!

PR #89, opened 2020-09-16
File transfer

piegamesde:
This is a manual copy-paste implementation of @vu3rdd's git branch (that got hopelessly outdated). The result was tweaked until the compiler is happy, but no further checks of functionality were performed.

General notice: I have no clue what I'm doing so please tell me what to do next ^^. Maybe could you please first check if I ported all the code correctly without missing any parts or introducing new bugs?


vu3rdd:
Thanks @piegamesde for all the hard work. Sorry for the late reply. I promise to have a look by this weekend.

piegamesde:
@vu3rdd The code is now finally "working". The transit handshake runs through and the file transfer also more or less works. Obviously, there are still known and unknown bugs by the dozen and at least as many `TODO`s. But I think now might be a good time to thing about how to mold this into a proper, usable API.

piegamesde:
Obsoleted by #94.

PR #103, opened 2021-01-13
CI: Improvements

saibotk:
This patch, removes the now unused CircleCI config and adjusts the Github Actions workflows to only run on pushes / PRs to the master branch. Any other branch may be considered work in progress, and we would not need the extra CI steps each commit (blocks potential other CI jobs in the project from running etc.) on an unstable branch.

This also re-enables running audit checks on path constraints, as the documentation now states, that it is available for both the push/pull_request rules (https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths). With the additional schedule for a regular check anyways.


piegamesde:
Thank you very much.

PR #107, opened 2021-01-22
Add overwrite checks

### LINKED ISSUE #105: Check files before receiving
We don't want to override random existing files on the receiver's host system, so we should check against that. The easiest way to do this is to simply check and bail out.

https://github.com/magic-wormhole/magic-wormhole.rs/blob/2b6605631d574f8556944d19837a25908e42eb60/src/transfer.rs#L329-L333

However, I'd prefer the CLI to ask me whether I want to override the file or not. Erroring out means that the transfer fails and the sender will have to re-send it and to retransmit a new code, which is annoying.

brightly-salty:
Fixes #105

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -283,15 +283,15 @@ async fn receive(mut w: Wormhole, relay_server: &str) -> anyhow::Result<()> {
     )
     .await;
 
-    let overwrite = !req.filename.exists()
-        || ask_user(
-            format!(
-                "Override existing file {}?",
-                req.filename.display()
-            ),
+    let overwrite = if req.filename.exists() {
+        ask_user(
+            format!("Override existing file {}?", req.filename.display()),
             false,
         )
-        .await;
+        .await
+    } else {
+        false
+    };
 
     if answer {
         req.accept(overwrite).await
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -2,7 +2,7 @@ use clap::{
     crate_authors, crate_description, crate_name, crate_version, App, AppSettings, Arg, SubCommand,
 };
 use log::*;
-use magic_wormhole::{transfer, CodeProvider, Wormhole};
+use magic_wormhole::{transfer, util::ask_user, CodeProvider, Wormhole};
 use std::{path::Path, str};
 
 #[async_std::main]
@@ -275,37 +275,11 @@ async fn receive(mut w: Wormhole, relay_server: &str) -> anyhow::Result<()> {
 
     let req = transfer::request_file(&mut w, &relay_server.parse().unwrap()).await?;
 
-    let mut stdout = io::stdout();
-    let stdin = io::stdin();
-
-    let answer = loop {
-        stdout
-            .write_fmt(format_args!(
-                "Receive file '{}' (size: {} bytes)? (y/N) ",
-                req.filename.display(),
-                req.filesize
-            ))
-            .await
-            .unwrap();
-
-        stdout.flush().await.unwrap();
-
-        let mut answer = String::new();
-        stdin.read_line(&mut answer).await.unwrap();
-
-        match answer.chars().next() {
-            Some('y') | Some('Y') => break true,
-            Some('n') | Some('N') => break false,
-            _ => {
-                stdout
-                    .write_fmt(format_args!("Please type y or n!\n"))
-                    .await
-                    .unwrap();
-                stdout.flush().await.unwrap();
-                continue;
-            },
-        };
-    };
+    let answer = ask_user(format_args!(
+        "Receive file '{}' (size: {} bytes)? (y/N) ",
+        req.filename.display(),
+        req.filesize
+    ));
 
     if answer {
         req.accept().await
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -2,7 +2,7 @@ use clap::{
     crate_authors, crate_description, crate_name, crate_version, App, AppSettings, Arg, SubCommand,
 };
 use log::*;
-use magic_wormhole::{transfer, util::ask_user, CodeProvider, Wormhole};
+use magic_wormhole::{transfer, util, CodeProvider, Wormhole};
 use std::{path::Path, str};
 
 #[async_std::main]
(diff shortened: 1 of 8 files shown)


src/bin/main.rs
```diff
@@ -275,37 +275,11 @@ async fn receive(mut w: Wormhole, relay_server: &str) -> anyhow::Result<()> {
...
-                    .write_fmt(format_args!("Please type y or n!\n"))
-                    .await
-                    .unwrap();
-                stdout.flush().await.unwrap();
-                continue;
-            },
-        };
-    };
+    let answer = ask_user(format_args!(
+        "Receive file '{}' (size: {} bytes)? (y/N) ",
```
iyzana, line 279: With a prompt like `(y/N)` (with an uppercase N) i would expect that simply pressing enter means no, but that is not the case with the `ask_user` function.

brightly-salty: Probably a typo in this case.. Maybe the ask_user function can have a boolean parameter which determines the default answer?

iyzana: A good typo in this case :)
Sound good!

piegamesde (changes requested):
Thank you so far! I've only got a few minor requests for now. Two things that I'd like to see happen in the future (however not necessarily in this PR if you don't want to):

- Don't make the user interaction in the library itself, but instead expose the conflict in the API somehow. Then, applications can do their custom input dialog if they want to.
- Add a few tests that try to receive a file but it already exists etc.

P.S.: On GitHub, you mark pull requests as "draft" instead of prepending "WIP" to the title

brightly-salty:
Okay @piegamesde I've made the changes I'm open to implementing the other changes (exposing it in the API, and adding tests) but I will probably need more guidance/mentoring to do something like that. I can continue in this PR or we can merge it and open a new one for those changes, whichever you prefer.

piegamesde:
I've had a look at it and the tests are in a rather bad shape at the moment, so doing it here would probably be scope creep. Regarding the API, I propose the following: The check is moved into `main.rs` and `accept` takes in a `overwrite: bool`. This allows for the following behavior:

- The application will check the file if it already exists and then pass the user's answer. If it doesn't, it will pass `false`. This should prevent any TOCTTOU bugs.
- One can safely always say `false` and never overwrite anything (for a future noninteractive mode)
- One can also pass `true` if the destination location is a temporary folder etc. (this will probably be useful for the tests as well)

brightly-salty:
@piegamesde So will the `RecieveRequest::accept()` function still check for overwriting? If it does, then we're duplicating effort in the application and the library.

piegamesde:
We have one explicit check and one implicit one when opening the file. If we move the check into the application, the library needs not do it anymore. The implicit check happens when we open the file: if `overwrite = false`, the `create_new` function will fail if the file already exists. If `overwrite = true`, the `truncate` option will handle existing files. By the way this still needs to be implemented, the starting point there is `let mut file = File::open` in `send_records`. I haven't tried it out myself, but your current code *should* fail to overwrite any files even when the user confirms it.

src/transfer.rs
```diff
@@ -8,10 +8,13 @@
 //! At its core, [`PeerMessage`s](PeerMessage) are exchanged over an established wormhole connection with the other side.
 //! They are used to set up a [transit] portal and to exchange a file offer/accept. Then, the file is transmitted over the transit relay.
 
+use async_std::fs::OpenOptions;
 use serde_derive::{Deserialize, Serialize};
 use serde_json::json;
 use std::sync::Arc;
 
+use crate::util::ask_user;
```
piegamesde, line 16: Unused import warning

src/bin/main.rs
```diff
@@ -271,44 +271,30 @@ async fn send_many(
...
-                stdout
-                    .write_fmt(format_args!("Please type y or n!\n"))
-                    .await
-                    .unwrap();
-                stdout.flush().await.unwrap();
-                continue;
-            },
-        };
-    };
+    let overwrite = !req.filename.exists()
```
piegamesde, line 286: This may be prone to race hazards (time to check vs time of use). The idea to resolve this is that if `!req.filename.exists()`, the user is not asked for confirmation, but `overwrite = false`. Therefore, a race hazard will not cause any files to unexpectedly be overwritten.

piegamesde (changes requested):
Two minor details and then this is good to go. As our CI is still WIP, please manually do the following things:

- Run `cargo +nightly fmt` for consistent code style
- Rewrite the history in pretty, otherwise I'll squash it into one commit if you're fine with that :)
- Tests
    - Send yourself a file
    - Send it again and check the correct behavior if you accept/deny to overwrite it.
- Regression tests
    - Send a file to yourself
    - Send it again, but using the python implementation as sender
    - And again, but using the python implementation as receiver

src/bin/main.rs
brightly-salty: Would this fix that? 
```rust
    let overwrite = if req.filename.exists() {
        ask_user(
            format!("Override existing file {}?", req.filename.display()),
            false,
        )
        .await
    } else {
        false
    };
```

piegamesde: Yes

piegamesde:
I just tried it out and I got a panic: `thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TransitType { abilities_v1: [DirectTcpV1, RelayV1], hints_v1: [DirectTcpV1(DirectHint { priority: 0.0, hostname: "192.168.1.1", port: 40900 }), RelayV1(RelayHint { hints: [DirectHint { priority: 0.0, hostname: "transit.magic-wormhole.io", port: 4001 }] })] }', src/transfer.rs:326:64`

Also a small typo in the dialog: Override=>Overwrite.

src/transfer.rs
```diff
@@ -322,13 +323,13 @@ impl<'a> ReceiveRequest<'a> {
             .connector
             .follower_connect(
                 self.wormhole.key.derive_transit_key(&self.wormhole.appid),
-                Arc::try_unwrap(self.other_side_ttype).unwrap(),
+                Arc::try_unwrap(self.other_side_ttype.clone()).unwrap(),
```
piegamesde, line 326: This should not work, because you are cloning the reference counter instead of the inner value. I think you want something like `(*self.other_side_ttype).clone(),` instead.

piegamesde:
Thank you for your contribution. I hope you don't mind me going ahead on the last meters :)

PR #108, opened 2021-01-22
Add cargo-deny

### LINKED ISSUE #99: Add deny.toml
https://github.com/EmbarkStudios/cargo-deny

We want to know about:
- Copyleft-licensed dependencies
- Dependencies with security issues
- Unmaintained dependencies

and probably a lot of other things.

brightly-salty:
Fixes #99 

Allowed licenses are MIT, Apache-2.0, ISC, BSD-2-Clause, BSD-3-Clause, and MPL.
RUSTSEC-2018-0015 is allowed for now

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,7 +1,6 @@
 targets = []
 
 [advisories]
-db-path = "~/.cargo/advisory-db"
 db-urls = ["https://github.com/rustsec/advisory-db"]
 vulnerability = "deny"
 unmaintained = "warn"
@@ -11,10 +10,10 @@ ignore = ["RUSTSEC-2018-0015"]
 
 [licenses]
 unlicensed = "deny"
-allow = ["MIT", "Apache-2.0", "ISC", "BSD-2-Clause", "BSD-3-Clause", "MPL-2.0"]
+allow = ["BSD-2-Clause", "MPL-2.0"]
 deny = []
 copyleft = "deny"
-allow-osi-fsf-free = "neither"
+allow-osi-fsf-free = "both"
 default = "deny"
 confidence-threshold = 0.8
 exceptions = []
@@ -24,7 +23,7 @@ ignore = false
 registries = []
 
 [bans]
-multiple-versions = "deny"
+multiple-versions = "warn"
 wildcards = "allow"
 highlight = "all"
 allow = []
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,63 @@
+targets = []
+
+[advisories]
+db-path = "~/.cargo/advisory-db"
+db-urls = ["https://github.com/rustsec/advisory-db"]
+vulnerability = "deny"
+unmaintained = "warn"
+yanked = "warn"
+notice = "warn"
+ignore = ["RUSTSEC-2018-0015"]
+
+[licenses]
+unlicensed = "deny"
+allow = ["MIT", "Apache-2.0", "ISC", "BSD-2-Clause", "BSD-3-Clause", "MPL-2.0"]
+deny = []
+copyleft = "deny"
+allow-osi-fsf-free = "neither"
+default = "deny"
+confidence-threshold = 0.8
+exceptions = []
+
+[licenses.private]
+ignore = false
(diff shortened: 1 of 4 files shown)


deny.toml
```diff
@@ -0,0 +1,63 @@
...
+ignore = false
+registries = []
+
+[bans]
+multiple-versions = "deny"
+wildcards = "allow"
+highlight = "all"
+allow = []
+deny = []
+skip = [
```
piegamesde, line 32: What exactly are all those entries needed for? What about setting `multiple-versions = "warn"` instead?

deny.toml
```diff
@@ -0,0 +1,63 @@
...
+[advisories]
+db-path = "~/.cargo/advisory-db"
+db-urls = ["https://github.com/rustsec/advisory-db"]
+vulnerability = "deny"
+unmaintained = "warn"
+yanked = "warn"
+notice = "warn"
+ignore = ["RUSTSEC-2018-0015"]
+
+[licenses]
```
piegamesde, line 12: Will this allow "public-domain"-like licenses? And why not set `allow-osi-fsf-free = "allow"`?

piegamesde (changes requested):
Thank you very much so far!

saibotk (changes requested):

deny.toml
brightly-salty: We will have to allow public-domain licenses individually when they occur, or I can add them now if you have specific examples. If I set `allow-ofi-fsf-free` to be allow, then copyleft license like GPL will be allowed as well.

deny.toml
```diff
@@ -0,0 +1,63 @@
+targets = []
+
+[advisories]
+db-path = "~/.cargo/advisory-db"
```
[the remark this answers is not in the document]
brightly-salty: This is what was provided by `cargo deny init` as the default. If you've looked into it and think it can be safely removed, I will remove it.

deny.toml
brightly-salty: These entries were provided by `cargo deny init`. I can set `multiple-versions` to warn if you want me to.

deny.toml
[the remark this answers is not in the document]
piegamesde: I've tried it out. It's not only safe, but it also works better.

deny.toml
piegamesde: Ah, I see. I've looked some more at the documentation:

> 1. A license specified in the deny list is always rejected.
> 2. A license specified in the allow list is always accepted.
> 3. If the license is considered copyleft, the [licenses.copyleft] configuration determines its status
> 4. If the license is OSI Approved or FSF Free/Libre, the [licenses.allow-osi-fsf-free] configuration determines its status, if it is neither the check continues
> 5. If the license does not match any of the above criteria, the [licenses.default] configuration determines its status

If I read this correctly, the `copyleft` deny should take precedence over `allow-osi-fsf-free`.

brightly-salty: Okay, I've changed it but not comitted. Looks like MPL-2.0 is considered copyleft, so we might want to investigate that:
```
error[L001]: failed to satisfy license requirements
  ┌─ wepoll-sys 3.0.1 (registry+https://github.com/rust-lang/crates.io-index):4:12
  │
4 │ license = "MPL-2.0"
  │            ^^^^^^^
  │            │
  │            license expression retrieved via Cargo.toml `license`
  │            rejected: license is considered copyleft
  │
  = wepoll-sys v3.0.1
    └── polling v2.0.2
        └── async-io v1.3.1
            ├── async-global-executor v2.0.2
            │   └── async-std v1.9.0
            │       ├── async-tungstenite v0.12.0
            │       │   └── magic-wormhole v0.1.0
            │       └── magic-wormhole v0.1.0 (*)
            └── async-std v1.9.0 (*)
```

piegamesde: I'd put MPL and maybe (big maybe) also LGPL under `allow` and then it's fine.

piegamesde (approved):

Issue #98, opened 2021-01-11
[transfer] Give progress information


piegamesde:
In order for the CLI to display a fancy progress bar, the transfer API needs to expose such information. Together with progress, we probably want to track transfer speed.

One possibly way to implement this is to pass some callback to the file transfer that calls the method once something happens. The alternative way would be to spawn the file transfer (or the UI updating) in a new task and let them send events over a channel. Note that the latter can be implemented using the former outside of the API (by having the callback writing into a channel).

The difficulty I see here is to know when the caller should be informed. After every packet? Every percent? Every second? Probably some combination of multiple. Can the transfer expect the UI updating to be reasonably quick? Otherwise we a forced to make this multiple tasks.

**Prior art/related work:**

- [progress-streams crate](https://docs.rs/progress-streams/1.1.0/progress_streams/): Wraps a `Reader`/`Writer` using the callback approach. Not sure if this fits us as we are on a higher abstraction level.

-----

The actual CLI implementation is secondary to this issue. Of course, it make sense to develop both together in order to test the ergonomics of the API, but we can always make the output pretty later on. There are plenty of fancy progress bar crates out there :)

brightly-salty:
Just wanted to put indicatif out there as a possible crate to help with this. You can provide it a callback and an optional theme and it will do the behind-the-scenes.
https://docs.rs/indicatif/0.15.0/indicatif/

moaz-mokhtar:
This is 2nd time to me to contribute in Rust project. Can I do a try?
I did a research and tried but found below:
1. progress-streams: have errors when implementing the sample code in the doc. That is why I preferred to check alternatives.
2. pbr: I found this crate which defined as: *Console progress bar for Rust Inspired from pb, support and tested on MacOS, Linux and Windows.*
    - Sample:
        ![image](https://user-images.githubusercontent.com/5870208/107549999-2a7f6780-6bd9-11eb-84a4-18971ae707c5.png)

piegamesde:
Sure, go ahead :)

I haven't tried `progress-streams` myself so I can't judge on it. Have you already found any alternatives?

I'd like to keep the scope of this issue small and mainly focus on the library part of it. For now, a few print statements on the application side will be fine, just enough to see that it's working.

moaz-mokhtar:
The alternate will be [pbr](https://crates.io/crates/pbr).

Regarding the scope, I understand that I have to do below:
- Show transfer data progress in the console.

[...]

PR #112, opened 2021-02-15
init_add_porgressbar

moaz-mokhtar:
Progress Bar added as discussed in issue #98 .

- I did use [pbr](https://crates.io/crates/pbr) library.
- After file sent and received, progress bar can be replaced with a message as below code, but I didn't apply it yet.
`pb.finish_print("Receive done");`
> bp: is the PorgressBar struct.
> finish_print() a function to show a message under impl of ProgressBar.


ryanmcgrath:
I think, ideally, this shouldn't use `ProgressBar` outside of the `bin/src` example - by using it in `src/transfer` you're expecting that every client would be a console application, but this doesn't hold true if someone's building a GUI on top of this.

moaz-mokhtar:
> I think, ideally, this shouldn't use `ProgressBar` outside of the `bin/src` example - by using it in `src/transfer` you're expecting that every client would be a console application, but this doesn't hold true if someone's building a GUI on top of this.

I'm very thankful for this notice @ryanmcgrath .
I'm going to redesign the solution but where should the console progress bar appeared?

ryanmcgrath:
So you want to edit `src/bin/main.rs` to do any actual `ProgressBar` creation/update calls. `src/transfer.rs` (and any other files) should really just have a reporting mechanism so that any higher level progress bar can reflect it.

You could expose this with a channel, I think - or maybe just take a callback method like `Fn(progress)` that fires, and any handler can then update however necessary.

moaz-mokhtar:
I have a challenge for this issue. Still I'm studying how to fix, but need more hints.

moaz-mokhtar:
I found a library called [notify](https://docs.rs/notify) which can notify about file status which can be used. But it is challenging how to apply the concept.

ryanmcgrath:
Mmm, I'm not sure this needs another crate to solve the issue - can you pass an `Fn(u64)` down, and instead of calling `pbr.set`, call that?

Then, for example, the `bin/` example could create a progressbar, and just pass a callback to update it.

piegamesde:
What exactly are your struggling with?

The way I'd go about this is to simply move the progress bar code into main.rs, and then ask myself "which information do I need that must be exposed by the library in order to make this work again?".

moaz-mokhtar:
Thank you for cooperation and feedback.

I did **implement** below:
#### transfer.rs 
```rust
// ============= Applying transfer progresison report ======
#[derive(Debug)]
pub struct TransferWatcher {
    transferred: u64,
    total: u64,
}

impl TransferWatcher {
    pub fn new(_total: u64) -> Self {
        TransferWatcher {
            total: _total,
            transferred: 0,
        }
    }

    pub fn update_transferred(&mut self, _transferred: u64) {
        self.transferred = self.transferred + _transferred;
    }
}

 ... 
// encrypt and send the file to tcp stream and return the sha256 sum
// of the file before encryption.
async fn send_records( ... ) -> Result<Vec<u8>> {
 ... 
    let mut transfer_watcher = TransferWatcher::new(file_size);
 ... 
    loop {
        // read a block of 4096 bytes
        let n = file.read(&mut plaintext[..]).await?;
        count += n;
        // debug!("sending {} bytes", n);

        // send the encrypted record
        transit.send_record(&plaintext[0..n]).await?;

        transfer_watcher.update_transferred(
            count
                .to_value()
                .to_u64()
                .expect("Can't convert to usize to u64."),
        );

 ... 
    }
    // pb.finish_print("Transfer done");

    Ok(hasher.finalize_fixed().to_vec())
}
```
---
My challenge is how the main.rs reported each time transferred data changed.
Below is the function in the main.rs which sending the file.
#### main.rs: send_file function
```rust
 ... 
transfer::send_file(&mut wormhole, file, &relay_server.parse().unwrap())
            .await
            .unwrap();
 ... 
```

#### transfer.rs: send_file function
```rust
/// Send a file to the other side
pub async fn send_file( ... ) -> Result<()> {
 ... 
    debug!("Beginning file transfer");

    tcp_file_send(&mut transit, &filepath)
        .await
        .context("Could not send file")
}
```

#### transfer.rs: tcp_file_send function
```rust
async fn tcp_file_send( ... ) -> Result<()> {
    // 11. send the file as encrypted records.
    let checksum = send_records(filepath, transit).await?;
 ... 
    debug!("transfer complete!");
    Ok(())
}

```

piegamesde:
Yes, that's a tricky question. You basically have two options on how to implement this right now:

1. You can take in a callback (as function pointer, probably `FnMut`) while sending, that you call every time some progress is made. The main then passes a callback that updates the progress bar from that data.
2. You can use a communication channel (`Sender`/`Receiver`) pair. The main then spawns a new task before sending, that listens in the loop on the receiver for some progress information to display. The transfer logic pushes updates on to the sender side when progress is made.

moaz-mokhtar:
Apologize, I'm still facing a challenge and working on it.

piegamesde:
I'm going to close this as superseded by #116. Thank you for your efforts. If you want, you can pick it up after #116 is merged and make the CLI more pretty.

moaz-mokhtar:
Very thankful for your cooperation, and apologize which I didn't finalize it.

piegamesde:
No worries. We're all doing it in our free time

PR #116, opened 2021-03-07
Adds progress reporting to transfers.

ryanmcgrath:
- Sending and Receiving (the latter via `accept()`) now accept a
  callback parameter for progress reporting, in the form of an
  `FnMut(u64, u64)`. The first parameter is the amount sent/received,
  the latter is the total.

- `bin/main.rs` is updated to implement a terminal progress bar, similar
  to pull request #112 from @moaz-mokhtar.

- Tests updated.

Happy to update if there's a preferred way to handle this, but it works well for me (both in the `bin` example, and my GUI work).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,13 +43,14 @@ pin-project = "1.0.4"
 # for "bin" feature
 clap = { version = "2.33.3", optional = true }
 env_logger = { version = "0.8.2", optional = true }
+pbr = { version = "1.0.4", optional = true }
 
 # for some tests
 [dev-dependencies]
 env_logger = "0.8.2"
 
 [features]
-bin = ["clap", "env_logger"]
+bin = ["clap", "env_logger", "pbr"]
 # TODO remove this one day
 # - Removing it now requires all cargo calls to have --features=bin which is annoying
 # - There is a cargo issue that would allow proper bin dependencies and thus would resolve it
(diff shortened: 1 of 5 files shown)


moaz-mokhtar:
Actually I'm happy and very thankful...
I will do a test in my local copy and enjoy it.

piegamesde:
Thank you. The result is not pretty, but that's okay – it's the API that matters. Please apply `cargo +nightly fmt`. You'll need to rebase on master for this if you're on the latest Rust version.

ryanmcgrath:
Cleaned this up and briefly looked into why the tests are failing on this, but offhand unsure why. Testing the actual binary (`bin/main.rs`) works fine for me, though. Lemme know if there's something I'm missing here?

piegamesde:
The current test situation is a mess, I know. You didn't break them, they were brittle to begin with and it was some subtle change that broke them (it's been a while, I don't remember, but maybe the nonce checking thing?). The whole test suite *badly* needs some love and an overhaul.

In the meantime, we're bound to manual testing. If sending and receiving work for both Rust and Python implementations, that's good enough for now (I'll probably test it tomorrow, but no promises).

ryanmcgrath:
Ah, yeah then - my tests go between a server of mine (Linux/Python) and my main machine (macOS/Rust). Send and receive across both with no issues.

PR #117, opened 2021-03-17
Refactor send-many

### LINKED ISSUE #95: Give `send-many` some love
At the moment it merely is a proof of concept. First of all, sends should be done in parallel. This means that once a connection to the other side is established, the actual file transfer should be moved into a background task in order for the main loop to accept new receivers.

The loop also needs an exit condition, based on time or number of receivers. I think 1h could be a sensible default.

tau3:
Hi! This PR closes #95


piegamesde:
Woah, that's awesome, many thanks!

Please be patient for this to get merged however, my backlog is a bit full these days

piegamesde:
Can you please rebase your changes on master?

tau3:
done. btw, looks like you have some broken unit tests in master (`core::test::test_file_rust2rust`)

piegamesde:
Yeah, I fucked up, see https://github.com/magic-wormhole/magic-wormhole.rs/pull/116#issuecomment-804493278

src/bin/main.rs
```diff
@@ -258,42 +266,44 @@ fn enter_code() -> anyhow::Result<String> {
...
+                            connector: WormholeConnector) -> anyhow::Result<()> {
+    let mut wormhole = connector.connect_to_client().await?;
+    task::spawn(async move {
+        let result = transfer::send_file(
+            &mut wormhole,
+            filename.deref(),
+            &url,
+            |sent, total| {
+                // @TODO: Not sure what kind of experience is best here.
+                info!("Sent {} of {} bytes", sent, total);
```
piegamesde, line 298: I think having an independent progress bar for each transfer would be best here. I haven't looked into how feasible this is though.

src/bin/main.rs
```diff
@@ -258,42 +266,44 @@ fn enter_code() -> anyhow::Result<String> {
...
-            },
-        }
+async fn send_many(relay_server: &str, code: &str, filename: &str) -> anyhow::Result<()> {
+    let time = SystemTime::now();
+    let one_hour = Duration::from_secs(3600);
+
+    let filename = Arc::new(filename.to_owned());
+    let url = Arc::new(relay_server.parse().map_err(|e| anyhow!("{}", e))?);
+
+    while time.elapsed()? < one_hour {
```
piegamesde, line 276: Why not use `Instant` instead of `SystemTime`? Its `elapsed()` is monotonic so it can't fail.

src/bin/main.rs
```diff
@@ -258,42 +266,44 @@ fn enter_code() -> anyhow::Result<String> {
...
-            Ok(_) => {
-                info!("TODO success message");
-            },
-            Err(e) => {
-                warn!("Send failed, {}", e);
-            },
-        }
+async fn send_many(relay_server: &str, code: &str, filename: &str) -> anyhow::Result<()> {
+    let time = SystemTime::now();
+    let one_hour = Duration::from_secs(3600);
```
piegamesde, line 271: Would be cool if you went the extra mile and made this a CLI parameter.

src/bin/main.rs
piegamesde: Also, I think there is some use case for limiting the attempts based on count as well. I'll have to calculate how many attempts are acceptable without having an increased risk of the communication being cracked. Maybe increase the default generated password length for `send-many` in order to counter this?

src/bin/main.rs
```diff
@@ -142,11 +150,11 @@ async fn main() -> anyhow::Result<()> {
                         .parse()
                         .expect("TODO error handling");
                     CodeProvider::AllocateCode(numwords)
-                },
+                }
```
piegamesde, line 153: I think I configured the formatter to use trailing commas. Maybe it's a nightly only feature, so try with `cargo +nightly fmt` and this change should go away.

piegamesde (changes requested):
Looks great so far

src/bin/main.rs
tau3: i'm not sure it can be answered without calculation :)

piegamesde:
I'm trying it out, and for some reason the help on that sub-command is empty? I can't tell any obvious cause for this though.

tau3:
it had been hidden by custom help message - fixed

piegamesde:
I somehow always get a 'crowded' error from the server. But I also get this on master. I suspect that it's the same bug that's causing the tests to fail at the moment. Unless you want to work on this some more, I think I'll merge this now and then fix it together with the tests in the (hopefully) near future.

Issue #123, opened 2021-04-10
Clipboard support


piegamesde:
It would be cool if the `send` command automatically copied the the code to the clipboard.

nsmlzl:
The CLI is supposed to run under Windows, macOS and Linux (X11 and Wayland), or? Any other platforms to consider?
Could we just use the [rust-clipboard](https://github.com/aweinstock314/rust-clipboard) crate? I did a quick test under X11 and Wayland; it seemed to work fine there.

Should the command be automatically copied? Or should we add a command line argument? E.g. [pass](https://git.zx2c4.com/password-store/about/) has a ```--clip``` argument.

Should we inform the user about the new data in the clipboard? Might be informative for new users, but would also add clutter to the output.

piegamesde:
I'd like to consider as many platforms as possible, of those where the application already runs. `rust-clipboard` looks fine, but it does not seem to have native Wayland support. This shouldn't be a problem as long as XWayland is a common thing, but we may want to look into this in the future.

> Should the command be automatically copied? Or should we add a command line argument? E.g. pass has a --clip argument.

That's a hard question. I asked a few people, and not all were enthusiastic about writing stuff to the clipboard. So if we make it a flag, should it be opt-in or opt-out? I fear that independently of that decision, half of the users will be unhappy about this. Maybe the best thing is to make this a persistent configuration option?

An alternative that I thought of would be "press some shortcut to copy the code to the clipboard". But I can't think of any good shortcuts that aren't captured by terminal emulators, so what do you think about "press enter to copy the code to the clipboard"?

> Should we inform the user about the new data in the clipboard?

Yes, absolutely, unless the action was explicitly requested by the user in the first place. I don't think that it would be cluttering the output.

corbinu:
I would make it off by default but maybe print a message informing people they can use it.

nsmlzl:
Is somebody already working on this issue? Or could I have a shot at it?

I was thinking about adding a ```--clip``` argument and also do a check for an environment variable (users could just add it to their bashrc to do a permanent configuration). Or does wormhole have its own configuration file? 

> An alternative that I thought of would be "press some shortcut to copy the code to the clipboard". But I can't think of any good shortcuts that aren't captured by terminal emulators, so what do you think about "press enter to copy the code to the clipboard"?

[...]

PR #134, opened 2021-10-29
Add option to pipe code to an external process

rpls:
This adds a command line flag `--command` for the `send{,-many}` commands that allows a user to specify a shell command to which the code-word is piped. This could, for example, enable a user to pipe the command to `xclip` to achieve #123. This is more versatile as this would allow a user to script some things, and probably a lot simpler than supporting clipboards in general.

It's just a crude first try, I'm not a rust programmer. But it works for me.


rpls:
Some questions:
- Should the command line flag be renamed?
- Should this use an asynchronous version of std::process (adds external dependency async-process)?

src/bin/main.rs
```diff
@@ -305,6 +313,7 @@ async fn parse_and_connect(
     let (wormhole, code) = match code {
         Some(code) => {
             if is_send {
+                execute_shell_command(term, matches, &code);
```
piegamesde, line 316: Maybe it would be better to make this option mutually exclusive with providing a code. Because if you provide a code to the application, you already know it.

src/bin/main.rs
```diff
@@ -51,6 +51,12 @@ async fn main() -> eyre::Result<()> {
         .takes_value(true)
         .value_name("FILE_NAME")
         .help("Suggest a different name to the receiver. They won't know the file's actual name on your disk.");
+    let shell_command = Arg::with_name("shell-command")
+        .long("command")
+        .short("s")
```
piegamesde, line 56: No short for this please

src/bin/main.rs
```diff
@@ -51,6 +51,12 @@ async fn main() -> eyre::Result<()> {
         .takes_value(true)
         .value_name("FILE_NAME")
         .help("Suggest a different name to the receiver. They won't know the file's actual name on your disk.");
+    let shell_command = Arg::with_name("shell-command")
+        .long("command")
```
piegamesde, line 55: Bike shed: how about `--handle-code`?

src/bin/main.rs
```diff
@@ -375,6 +385,27 @@ fn sender_print_code(term: &mut Term, code: &magic_wormhole::Code) -> eyre::Resu
     Ok(())
 }
 
+fn execute_shell_command(
+    term: &mut Term,
+    matches: &clap::ArgMatches<'_>,
+    code: &magic_wormhole::Code
+) -> eyre::Result<()> {
+    if let Some(shell_command) = matches.value_of_os("shell-command") {
+        let mut child = std::process::Command::new("/bin/sh")
```
piegamesde, line 394:
```suggestion
        let mut child = std::process::Command::new("sh")
```

Don't hard-code the absolute path. Alternatively, `/usr/bin/env sh` should work fine as well.

src/bin/main.rs
```diff
@@ -375,6 +385,27 @@ fn sender_print_code(term: &mut Term, code: &magic_wormhole::Code) -> eyre::Resu
...
+            .arg("-c")
+            .arg(shell_command)
+            .stdin(std::process::Stdio::piped())
+            .spawn()
+            .expect("failed execute shell-command");
```
piegamesde, line 399: Error handling instead of panic:

```suggestion
            .context("failed execute shell-command")?;
```

(same applies for the lines below too)

src/bin/main.rs
```diff
@@ -375,6 +385,27 @@ fn sender_print_code(term: &mut Term, code: &magic_wormhole::Code) -> eyre::Resu
...
+        let mut stdin = child.stdin.take().expect("failed to open stdin of shell-command");
+        stdin.write_all(code.as_bytes()).expect("failed to write to stdin of shell-command");
+        drop(stdin);
+        let status = child.wait().expect("failed to wait for shell-command exit");
```
piegamesde, line 403: Please check the exit and fail if it is non-zero (I don't think this is already happening here).

piegamesde (commented):
Thank you for the PR. It's a worth while feature to have right now, but be aware that clipboard support might make it obsolete one day.

------

> Should the command line flag be renamed?

Yes, see my review comment.

> Should this use an asynchronous version of std::process (adds external dependency async-process)?

To be honest, not sure. I think we can always change it to `async` if the need ever comes up.

piegamesde:
I just got an idea for a different approach to the problem that I'd like to discuss: how about having a command that generates a code? Example usage:

```sh
code=$(wormhole generate-code)
wormhole send --code=$code
wormhole recieve $code
```

```sh
code=$(wormhole generate-code)
xclip $code
wormhole send --code=$code
```

piegamesde:
Since when do we have a code coverage bot?! :D

src/bin/main.rs
rpls: I guess `Command` handles the path search, so just `sh` is fine. Should I also include a Windows if/else that uses `cmd /C`?

src/bin/main.rs
rpls: > I just got an idea for a different approach to the problem that I'd like to discuss: how about having a command that generates a code? Example usage:
> 
> ```shell
> code=$(wormhole generate-code)
> wormhole send --code=$code
> wormhole recieve $code
> ```
> 
> ```shell
> code=$(wormhole generate-code)
> xclip $code
> wormhole send --code=$code
> ```

That would work too I guess. And users could probably have a shell `alias` or function in their profile to have this in one go.

rpls:
> I just got an idea for a different approach to the problem that I'd like to discuss: how about having a command that generates a code? Example usage:
> 
> ```shell
> code=$(wormhole generate-code)
> wormhole send --code=$code
> wormhole recieve $code
> ```
> 
> ```shell
> code=$(wormhole generate-code)
> xclip $code
> wormhole send --code=$code
> ```

That would work too I guess. And users could probably have a shell alias or function in their profile to have this in one go.

src/bin/main.rs
piegamesde: Ugh, Windows support. To be honest I don't really care, so try to make it work, otherwise simply gate that option behind a feature flag (something like `#[cfg(not(target_os = "windows"))]` IIRC, but I recommend reading [the documentation on this](https://doc.rust-lang.org/reference/conditional-compilation.html)).

rpls:
I worked in all the suggestions, but the windows support is untested.

src/bin/main.rs
rpls: Had to include some things from the `eyre` namespace.

src/bin/main.rs
rpls: I added an appropriate check an generate an error if the RC is non-zero.

src/bin/main.rs
rpls: Sounds good!

zeri42:
> ```shell
> generate-code)
> wormhole send --code=$code
> wormhole recieve $code
> ```

if you do that you can use anything as a code that starts with a number like this:
```
code=$((RANDOM%1024))-$(head -c 16 /dev/urandom | xxd -ps)
xclip $code
wormhole send --code=$code
```

piegamesde:
I just made #141. Please give it a try. Do we still need your proposed solution for some use cases? Are those use cases worth supporting?

Issue #152, opened 2022-04-04
Sender hangs when the file size is a multiple of 4096


kanna5:
To reproduce:

```bash
dd if=/dev/zero of=test.bin bs=4096 count=123
wormhole-rs send test.bin
```

The receiver will exit once all the contents of the file have been received, but the sender will hang until `ctrl+c` is pressed twice.

aeshirey:
Newbie here interested in contributing.

It looks like the cause of this is in [`transfer::v1::send_records`](https://github.com/magic-wormhole/magic-wormhole.rs/blob/b6a471a67dd1158303f65aee3848ae6a5b72a48f/src/transfer/v1.rs?q=4096#L401) where the async read is performed. In the case of an input of size 4096, there's nothing left to read, so the sender somehow just spins. ¯\\\_(ツ)\_/¯

Fortunately, this function receives `file_size`, so it looks like this should be a straightforward fix:

```rust
if n < 4096 {
    break;
} else if sent_size == file_size {
    // multiple of 4096
    break;
}
```

I already verified that @kanna5's `dd` input hangs the sender and that this additional check fixes it. Happy to submit a PR. I'm curious why the current check doesn't simply do this, though?

piegamesde:
Thanks for having a look. I'm not sure about checking `sent_size == file_size`, since this may go bad if the file is modified concurrently (time of check vs time of use). We could do `if n < 4096 || sent_size >= file_size`, but this would still hang if the file was concurrently truncated to a multiple of 4096 bytes. This would be an acceptable compromise for me, but there really should be a better solution to this.

aeshirey:
Ah, you're right. In my rudimentary testing, it didn't appear to be the case:

```rust
eprintln!("Reading...");
let n = file.read(&mut plaintext[..]).await?;
eprintln!("Read {n} bytes");
```

"Read 0 bytes" never showed up, maybe the progress bar hampering stdout? But `assert!(n > 0)` crashes.

PR #153, opened 2022-04-08
Break when reading at EOF

aeshirey:
Fix for #152 - when reading zero bytes at EOF, don't transmit, update sent size, or update hash.


piegamesde:
Thanks. Can you please also add an assertion about the payload length to `Transit::send_record`? And if you want to go the extra mile, add a test case to catch any regressions (could probably be based off our `test_file_rust2rust` test).

aeshirey:
Binary file example-file-4096.bin created according to the original issue (`dd if=/dev/zero of=example-file-4096.bin bs=4096 count=123`).

All tests ran successfully (`cargo test --all`) and specifically, the new test (`cargo test test_4096_file_rust2rust --all`).

piegamesde:
Looking good! However, I'd like to not check in a 0.5 MiB file into the repository if I can avoid it. I think something smaller, like 8096 bytes should still do the trick.

(Also, rustfmt is having a fit for some reason right now, just ignore it)

PR #163, opened 2022-05-31
Add WASM library code

andipabst:
This PR adds the library code from https://github.com/andipabst/magic-wormhole-wasm for wasm support. 
Related to #2 and #161

### DIFF
--- a/wasm/src/lib.rs
+++ b/wasm/src/lib.rs
@@ -1,0 +1,180 @@
+use std::{error::Error, path::PathBuf, rc::Rc};
+
+use futures_util::future::FutureExt;
+use log::Level;
+use magic_wormhole::{transfer, transit, Code, Wormhole};
+use wasm_bindgen::prelude::*;
+
+mod event;
+
+#[cfg(feature = "wee_alloc")]
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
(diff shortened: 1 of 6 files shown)


andipabst:
Since there are no checks on PRs to branches other than master, you can also look at my fork for the GitHub Actions results: https://github.com/andipabst/magic-wormhole.rs/pull/1

wasm/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+/// Send a file via wormhole.
+/// * `config` - The configuration object for the wormhole, created by [WormholeConfig::new]
+/// * `file` - The Blob with the file to send
+/// * `file_name` - The name of the file to send
+/// * `cancel` - A promise that cancels the transfer, if it is resolved
+/// * `event_handler` - A callback function that receives [event::Event] objects with updates about
+/// the sending status and progress
+pub async fn send_file(
+    config: WormholeConfig,
+    file: web_sys::Blob,
```
piegamesde, line 57: Obviously the API does not expose everything from the Rust API yet, but one important thing would be bytes streaming so that the file does not need to be held in memory in its entirety.

piegamesde, line 57: Obviously the API does not expose everything from the Rust API yet, but one important thing would be bytes streaming so that the file does not need to be held in memory in its entirety.

PR #164, opened 2022-06-15
Replace unmaintained get_if_addrs crate with if-addrs

felinira:
This replaces the unmaintained `get_if_addrs` crate with a currently maintained fork. This gets rid of a few old dependencies as a side effect as well.

I tested running a transfer and it seems to work as usual.


piegamesde:
It looks like rustfmt changed in the meantime, please run `cargo +nightly fmt` and put the changes in a separate commit since they're independent.

PR #171, opened 2022-10-14
Use stable tagged cli-clipboard dependency

alerque:
Upstream finally tagged a release that includes the fixes that were
forcing a Git HEAD usage here. In order for distro packages to be
reproducible it is important to use tagged dependency versions,
especially at tagged release points.

https://github.com/ActuallyAllie/cli-clipboard/issues/13

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -376,8 +376,9 @@ dependencies = [
 
 [[package]]
 name = "cli-clipboard"
-version = "0.2.1"
-source = "git+https://github.com/ActuallyAllie/cli-clipboard#c376ba6ce17b71feeed1693f1e5267cf73d7d14b"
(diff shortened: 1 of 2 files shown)


alerque:
The failing CI check for cargo fmt has nothing to do with this PR.

piegamesde:
Yeah, I'm honestly annoyed at `cargo fmt` for a while now. In one PR it requests a change, in the next one it wants that change back. The problem is that we are depending on some feature only available on nightly, so it is inherently unstable …

piegamesde:
Change has been merged, even though GitHub has not gotten the memo yet (tbf I did some rebasing and cherry-picking, but I don't remember having changed the commit itself).

PR #173, opened 2022-11-08
Add subcommand for generating shell completions

### LINKED ISSUE #146: Generate shell completion
They should come from our CLI arguments parser, `clap`. They very likely provide some tooling for this.

Stupremee:
Resolves #146

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -7,7 +7,7 @@ use std::{
 };
 
 use async_std::{fs::OpenOptions, sync::Arc};
-use clap::{Args, Parser, Subcommand};
+use clap::{Args, CommandFactory, Parser, Subcommand};
 use cli_clipboard::{ClipboardContext, ClipboardProvider};
 use color_eyre::{eyre, eyre::Context};
 use console::{style, Term};
@@ -225,6 +225,12 @@ enum WormholeCommand {
     /// Forward ports from one machine to another
     #[clap(subcommand)]
     Forward(ForwardCommand),
+    /// Generate shell completions for the wormhole CLI
+    #[clap(hide = true)]
+    Completion {
+        /// The shell type to generate completions for (bash, elvish, powershell, zsh)
+        shell: clap_complete::Shell,
+    },
     #[clap(hide = true)]
     Help,
 }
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -551,6 +557,35 @@ async fn main() -> eyre::Result<()> {
...
+                    let mut out = Vec::new();
+                    clap_complete::generate(shell, &mut cmd, &bin, &mut out);
+
+                    let out = match String::from_utf8(out) {
+                        Ok(x) => format!(
+                            "compdef _{0} {0}\n _{0}() {{ {1} }}\n\nif [ \"$funcstack[1]\" = \"{0}\" ]; then\n   {0} \"$@\"\nfi",
+                            bin, x
+                        )
+                        .into_bytes(),
+                        Err(out) => out.into_bytes(),
```
piegamesde, line 578: That's a bit weird, why would we silently not process the output if it is not valid UTF-8? I think requiring UTF-8 is valid here, so you can just propagate the error up and then do business as usual.

cli/src/main.rs
```diff
@@ -226,6 +227,11 @@ enum WormholeCommand {
     /// Forward ports from one machine to another
     #[clap(subcommand)]
     Forward(ForwardCommand),
+    /// Generate shell completions for the wormhole CLI
+    Completion {
+        /// The shell type to generate completions for (bash, elvish, powershell, zsh)
+        shell: Shell,
```
piegamesde, line 233: That line got me confused in the first read, please use `clap_complete:Shell` instead of the import syntax to make it easier to see that it is coming from a different crate.

cli/src/main.rs
```diff
@@ -226,6 +227,11 @@ enum WormholeCommand {
     /// Forward ports from one machine to another
     #[clap(subcommand)]
     Forward(ForwardCommand),
+    /// Generate shell completions for the wormhole CLI
+    Completion {
```
piegamesde, line 231: I would not expect end users to ever use this command (completions should be provided by the package already). Therefore we can hide it from the help output

piegamesde (changes requested):
Thank you, looks reasonable so far. I left a few small comments

PR #176, opened 2022-12-02
Send client version with bind message

wuan:
This PR sets the `client_version` field in the `bind` message (see [magic-wormhole/magic-wormhole/src/wormhole_rendezvous.py:182](https://github.com/magic-wormhole/magic-wormhole/blob/f372ccc466a4f1a909fae522d54f13f531d529f0/src/wormhole/_rendezvous.py#L182)). So far this is only being used by the Python and Wormhole-William (Go) clients.

It will use the following mapping:

|      |       |
|-----|-----|
| implementation | `rust` |
| version | `<package-version>` |

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -477,26 +477,15 @@ impl<S: Into<String>> From<S> for EitherSide {
 }
 
 #[derive(PartialEq, Eq, Clone, Debug, Deserialize, Serialize, derive_more::Display)]
-#[display(fmt = "{}-{}", "&*_0[0]", "&*_0[1]")]
-pub struct ClientVersion<'a>(#[serde(borrow)] pub [&'a str; 2]);
+#[display(fmt = "{}-{}", "&*_0", "&*_1")]
+pub struct ClientVersion(pub Cow<'static, str>, pub Cow<'static, str>);
 
-impl<'a> ClientVersion<'a> {
-    pub fn new(name: &'a str, version: &'a str) -> Self {
-        let array = [name, version];
-        ClientVersion { 0: array }
-    }
-}
-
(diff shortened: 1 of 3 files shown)


piegamesde:
This field does not seem to be specified in the protocols. We should fix that

piegamesde:
Using lifetimes here has the effect of avoiding about two allocations in total and is total overkill. But if you really want to have a borrowed version of the struct, then use `Cow` instead please.

wuan:
I'm new to Rust and need to find out how to avoid lifetimes then.

piegamesde:
Just clone liberally. And if you find yourself cloning too much, wrap in a reference counter to make it cheaper.

One of the key lessons most beginners eventually learn is that Rust gives you the *possibility* to be efficient, but it's not always necessary/productive to make use of that. (Incidentally, I think this is also where a lot of the "fighting the borrow checker" meme comes from)

wuan:
The solution is now hopefully clean enough.

piegamesde:
Yes, it is.

Unless this PR is urgent, I'd like to discuss the `client-version` field in the protocol before merging this. Depending on what the values mean, I'd personally prefer `CLIENT_NAME = "wormhole-rs";` over `"rust"`. On the other hand, there may exist multiple clients using this library, should the value reflect that or not?

wuan:
> On the other hand, there may exist multiple clients using this library, should the value reflect that or not?

That would be a future update to enable overriding this information by the caller, like in https://github.com/psanford/wormhole-william/blob/master/rendezvous/options.go.

piegamesde:
https://github.com/magic-wormhole/magic-wormhole-protocols/issues/35

Issue #183, opened 2023-02-28
Add function to get claimed nameplates


JustusFT:
It would be nice to have a function that returns a list of claimed nameplates. This way, when a user tries to receive something using a code with an unclaimed nameplate, clients could add their own error handling.

PR #184, opened 2023-02-28
Add nameplates fn

JustusFT:
Ref: https://github.com/magic-wormhole/magic-wormhole.rs/issues/183

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -42,6 +42,8 @@ pub enum WormholeError {
     PakeFailed,
     #[error("Cannot decrypt a received message")]
     Crypto,
+    #[error("Nameplate is unclaimed: {}", _0)]
+    UnclaimedNameplate(Nameplate),
 }
 
 impl WormholeError {
@@ -165,6 +167,7 @@ impl Wormhole {
     pub async fn connect_with_code(
         config: AppConfig<impl serde::Serialize + Send + Sync + 'static>,
         code: Code,
+        expect_claimed_nameplate: bool,
     ) -> Result<(WormholeWelcome, Self), WormholeError> {
         let AppConfig {
             id: appid,
@@ -174,6 +177,15 @@ impl Wormhole {
         let (mut server, welcome) = RendezvousServer::connect(&appid, &rendezvous_url).await?;
 
         let nameplate = code.nameplate();
+
+        if expect_claimed_nameplate {
+            let nameplate = code.nameplate();
+            let nameplates = server.list_nameplates().await?;
+            if !nameplates.contains(&nameplate) {
+                return Err(WormholeError::UnclaimedNameplate(nameplate));
+            }
+        }
+
         let mailbox = server.claim_open(nameplate).await?;
         log::debug!("Connected to mailbox {}", mailbox);
 
--- a/src/core/rendezvous.rs
+++ b/src/core/rendezvous.rs
@@ -74,6 +74,10 @@ impl RendezvousError {
 
 type MessageQueue = VecDeque<EncryptedMessage>;
 
+#[derive(Clone, Debug, derive_more::Display)]
+#[display(fmt = "{:?}", _0)]
+struct NameplateList(Vec<Nameplate>);
+
 #[cfg(not(target_family = "wasm"))]
 struct WsConnection {
     connection: async_tungstenite::WebSocketStream<async_tungstenite::async_std::ConnectStream>,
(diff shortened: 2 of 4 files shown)


src/core/rendezvous.rs
```diff
@@ -528,6 +541,15 @@ impl RendezvousServer {
             .is_some()
     }
 
+    pub async fn nameplates(&mut self) -> Result<Vec<Nameplate>, RendezvousError> {
```
piegamesde, line 544:
```suggestion
    pub async fn list_nameplates(&mut self) -> Result<Vec<Nameplate>, RendezvousError> {
```

Also add a short doc comment about what this does and when it can be used.

Is the usage of this command restricted in any way by the state machines? As in, is it even allowed to execute it after having claimed a nameplate? Is it allowed after the nameplate was released?

piegamesde (changes requested):
Looking okay so far. One thing I'm unsure about is any public API: as proposed here, you'd need access to the `RendezvousServer` in order to make the call. As far as I can tell, it is only exposed when using `Wormhole::connect_custom`.

I don't know what the intended use cases are\*, so maybe this is all the necessary API.

\* autocompletion is [out](https://github.com/magic-wormhole/magic-wormhole/issues/471)

src/core/rendezvous.rs
```diff
@@ -74,6 +74,15 @@ impl RendezvousError {
 
 type MessageQueue = VecDeque<EncryptedMessage>;
 
+#[derive(Clone, Debug)]
+pub struct NameplateList(Vec<Nameplate>);
+
+impl std::fmt::Display for NameplateList {
```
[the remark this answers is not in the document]
JustusFT: The compiler says the macro can't be applied on `Vec<T>` types, so I made the impl manually
I get this error:
```
error[E0277]: `Vec<Nameplate>` doesn't implement `std::fmt::Display`
  --> src/core/rendezvous.rs:77:24
   |
77 | #[derive(Clone, Debug, derive_more::Display)]
   |                        ^^^^^^^^^^^^^^^^^^^^ `Vec<Nameplate>` cannot be formatted with the default formatter
   |
   = help: the trait `std::fmt::Display` is not implemented for `Vec<Nameplate>`
   = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
   = note: this error originates in the derive macro `derive_more::Display` (in Nightly builds, run with -Z macro-backtrace for more info)

```

JustusFT:
I updated the linked issue with more information about the use case.
Currently I want the application to exit if the receiver put in a code with an unclaimed nameplate.
Would it be better to include the nameplate validation inside `connect_with_code` so we don't need to call `connect_custom`? This would be similar to how wormhole-william does it

piegamesde:
That would be okay to do, however it should be opt-in, as in most scenarios you actually don't want such a check.

src/core/rendezvous.rs
JustusFT: I updated the function to have a doc comment. Let me know if the wording is okay

It can be called at any time. (before claiming a namplate, after claiming a nameplate, and after releasing a nameplate)

src/core.rs
```diff
@@ -174,6 +177,15 @@ impl Wormhole {
         let (mut server, welcome) = RendezvousServer::connect(&appid, &rendezvous_url).await?;
 
         let nameplate = code.nameplate();
+
+        if expect_claimed_nameplate {
+            let nameplate = code.nameplate();
+            let nameplates = server.list_nameplates().await?;
+            if !nameplates.iter().any(|e| e.0 == nameplate.0) {
```
piegamesde, line 184: Something more like

```suggestion
            if !nameplates.contains(&nameplate)
```

src/core/rendezvous.rs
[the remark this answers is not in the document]
piegamesde: Try in combination with `#[display(fmt = "{:?}", _0)]`

src/core/test.rs
```diff
@@ -75,7 +75,8 @@ pub async fn test_file_rust2rust() -> eyre::Result<()> {
             let code = code_rx.await?;
             log::info!("Got code over local: {}", &code);
             let (welcome, wormhole) =
-                Wormhole::connect_with_code(transfer::APP_CONFIG.id(TEST_APPID), code).await?;
+                Wormhole::connect_with_code(transfer::APP_CONFIG.id(TEST_APPID), code, false)
```
piegamesde, line 78: Let's set this to true where possible to also test the happy case. (Also in the other tests)

piegamesde (changes requested):

piegamesde (approved):
Please make the commit history pretty (squashing everything together is okay too) and then we are good to go.

JustusFT:
@piegamesde I squashed and rebased :+1:

Issue #131, opened 2021-10-02
Support text messages


piegamesde:
The protocol level on that one is rather easy. Implementation notes:

- **Sender side:**
  - Add a `--text` file to the send command
  - If the file to be sent is `-`, the content from `stdin` is sent as text message. Interactive prompts are disabled (though I don't think we have any at the moment).
  - Don't forget to limit the input size to something sensible.
- **Receiver side:**
  - Text messages are always accepted by default
  - Do some quick content type checking and maybe display binary files differently
  - Maybe filter terminal escape codes? (Also maybe sending colored text is a feature, but are there any security risks?)
  - Check text size again before printing (maybe automatically open `$PAGER` with the content on long messages?)
  - Add CLI flag to write the message to a file instead (optional IMO)

felinira:
I would be interested in being able to receive text messages that were sent from the python client. Once in a while someone tries this and is presented with an unhelpful error message. This would probably be easiest to implement on top of the transfer-v2 API change once that is done, right?

Downstream issue: https://gitlab.gnome.org/World/warp/-/issues/30

j-lakeman:
Any updates on this?
This feature would be really handy IMHO to send secrets e.g. passwords.

PR #185, opened 2023-03-07
add timeout to clipboard initialization

jonZlotnik:
**TL;DR**
- `cli_clipboard` hangs on initialization in some weird x11 setups (e.g. my WSL2)
- In linux, `cli_clipboard` depends on a connection to a socket.
- the clipboard functionality isn't critical, so we shouldn't fail if not working

**Context:**
I wanted to try my hand at the [sending text](https://github.com/magic-wormhole/magic-wormhole.rs/issues/131 ) feature,
so I tried getting this repo setup on WSL2, but I couldn't get my binary to run. It kept hanging with no output.

**Into debug mode I went:**
All I got was `[2023-03-06T23:58:31Z DEBUG wormhole_rs] Logging enabled.`

**Now to debug Rust for the first time:**
I found that it was stuck inside the cli_clipboard ClipboardContext::new() function.
And noticed that for Linux environments, it depends on an x11 socket.

image

There doesn't seem to be any kind of test suite for the cli at the moment, so let me know if there something I should add.

Note: I'm brand spanking new to Rust. So please be verbose with comments and teachings. 😅

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -3,6 +3,8 @@ mod util;
 
 use std::{
     ops::Deref,
+    sync::mpsc,
+    thread,
     time::{Duration, Instant},
 };
 
@@ -283,11 +285,31 @@ async fn main() -> eyre::Result<()> {
             .try_init()?;
     }
 
-    let mut clipboard = ClipboardContext::new()
-        .map_err(|err| {
-            log::warn!("Failed to initialize clipboard support: {}", err);
-        })
-        .ok();
+    // cli_clipboard can hang if unable to connect to x11 socket in linux
+    // so we try for up to 1 second
+    let (sender, receiver) = mpsc::channel();
+    let _clipboard_init_thread = thread::spawn(move || {
+        match sender.send(
+            ClipboardContext::new()
+                .map_err(|err| {
+                    log::warn!("Failed to initialize clipboard support: {}", err);
+                })
+                .ok(),
+        ) {
+            Ok(()) => {}, // everything good
+            Err(_) => {}, // we have been released, don't panic
+        }
+    });
+    let mut clipboard = match receiver.recv_timeout(Duration::from_millis(1000)) {
+        Ok(initialized_clipboard) => {
+            log::debug!("Clipboard initialized successfully");
+            initialized_clipboard
+        },
+        Err(err) => {
+            log::warn!("Clipboard initialization timed out: {}", err);
+            None
+        },
+    };
 
     let concat_file_name = |file_path: &Path, file_name: Option<_>| {
         // TODO this has gotten out of hand (it ugly)


cli/src/main.rs
```diff
@@ -283,11 +283,31 @@ async fn main() -> eyre::Result<()> {
...
+            ClipboardContext::new().map_err(|err| {
+                log::warn!("Failed to initialize clipboard support: {}", err);
+            })
+            .ok()
+        ){
+            Ok(()) => {}, // everything good
+            Err(_) => {}, // we have been released, don't panic
+        }
+    });
+    let mut clipboard = match receiver.recv_timeout(Duration::from_millis(1000)) {
```
jonZlotnik, line 300: 1 second is probably over-cautious, if anybody can justify a shorter timeout, pls do. (I didn't do any research into how long it takes for cli_clipboard to init normally)

.gitignore
```diff
@@ -6,3 +6,4 @@
 **/*.rs.bk
 
 .idea/
+.vscode/
```
jonZlotnik, line 9: I use vscode... others probably do to?

jonZlotnik:
Note: I'm actually having trouble replicating the issue I faced in other environments (even other WSL2 envs).
Nonetheless, blocking the main thread of any program on waiting for an "networky" OS resource without a timeout is usually not a good idea.

piegamesde:
I'll try to have a look at it in the next couple of days. In the mean time, would you mind reporting the bug upstream to the relevant crate? Maybe they have some deadlock or other bug that can be fixed for that platform.

jonZlotnik:
Issues seem to already be stacking up. Maybe switching to 1Password's arboard https://github.com/1Password/arboard could be good?

cli_clipboard
https://github.com/ActuallyAllie/cli-clipboard/issues/16
https://github.com/ActuallyAllie/cli-clipboard/issues/17

x11_clipboard
https://github.com/quininer/x11-clipboard/issues/32

@piegamesde you also added this and it's still open
https://github.com/ActuallyAllie/cli-clipboard/issues/10

piegamesde:
Thanks for digging into this. I'm fine if you want to make a switch to arboard, the only requirement I have is that it has native Wayland support. Using it would have the added benefit of not depending on any native X libraries, as we currently do.

piegamesde:
Finally had a look at the code. It looks okay, but I think it can be simplified: since we always only send one value, you can simply have it as return value of the closure and not use a channel at all. On the main thread, one would then `join` the clipboard thread with a timeout. If it succeeds print the value, otherwise kill the thread and print an error message.

Of course you could also try out arboard, which will probably make the background thread obsolete.

jonZlotnik:
Okay, so I think you're correct in that the timeout fix should be upstream, but the cli_clipoard is too far downstream (depends on x11_clipboard, which depends on x11rb).

So I tried switching to arboard, but it actually has the same issue. Thankfully, they depend on x11rb directly, and a small fix is possible. So I submitted a PR there, and we'll see how it progresses.

https://github.com/1Password/arboard/pull/100

jonZlotnik:
> Finally had a look at the code. It looks okay, but I think it can be simplified: since we always only send one value, you can simply have it as return value of the closure and not use a channel at all. On the main thread, one would then `join` the clipboard thread with a timeout. If it succeeds print the value, otherwise kill the thread and print an error message.
> 
> Of course you could also try out arboard, which will probably make the background thread obsolete.

I actually don't think it's possible without a channel unless there's some API I missed for joining threads. (very well might be)
Don't want to wait the full time if the thread returns successfully.

jonZlotnik:
> Thanks for digging into this. I'm fine if you want to make a switch to arboard, the only requirement I have is that it has native Wayland support. Using it would have the added benefit of not depending on any native X libraries, as we currently do.

Alright, so my PR finally got merged in https://github.com/1Password/arboard/pull/100

There seems to be the option to have Wayland by default:
https://github.com/1Password/arboard/blob/3399a2582f33085f9602b1acfaaa5847face8373/README.md?plain=1#L21-L27

piegamesde:
Okay, how to proceed with this? Do you want to try switching to arboard?

jonZlotnik:
> Okay, how to proceed with this? Do you want to try switching to arboard?

Yeah I'll PR the switch to arboard. Sorry for the lapse.

jonZlotnik:
Switch is PR'd in #203

jonZlotnik:
Switch is PR'd in https://github.com/magic-wormhole/magic-wormhole.rs/pull/203

PR #186, opened 2023-03-07
fix tarpaulin code coverage and add .vscode to gitignore

jonZlotnik:
the github action for tarpauling breaks often, and it's not worth depending on it when it's just as easy (and more transparent) to run it using the cargo action

evidence of frequent breakage:
https://github.com/actions-rs/tarpaulin/issues/15
https://github.com/actions-rs/tarpaulin/pull/23


jonZlotnik:
wow... I'm terrible at git

PR #190, opened 2023-04-17
Unify sending/receiving progress bar style

afontenot:
Change the receiving end progress bar so that, like the sending end, it will only be updated every 250 ms. This avoids extremely rapid flickering that makes the ETA very hard to read.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -702,6 +702,17 @@ fn create_progress_bar(file_size: u64) -> ProgressBar {
     pb
 }
 
+fn create_progress_handler(pb: ProgressBar) -> impl FnMut(u64, u64) {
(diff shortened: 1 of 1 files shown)


afontenot:
Fixed - I think the test failure is bogus.

piegamesde:
Yes, some tests are flaky

piegamesde:
The reason this bug has happened is that all progress bar handlers are identical, but the code is duplicated all over the place. So I probably just missed that one the last time I touched the code. To prevent this from happening in the future, would you mind factoring them out into their own function? I'm thinking about something with the following signature: `fn progress_handler(pb: ProgressBar) -> impl FnMut(u64, u64)`

Issue #106, opened 2021-01-15
New wordlists


piegamesde:
Since I've started using Magic Wormhole, I've never been fond of the word list it uses. Quick summary:

- It's called the [PGP word list](https://en.wikipedia.org/wiki/PGP_word_list)
- It contains two sets of 256 words which are used in alternance (one byte per word)
- The words are chosen to be phonetically distinct for oral transmission.

Here's the problems I have with it:

- The phonetic distinction and the even/odd thing only work if you already know the word list by heart. The words may still be similar to others that are *not* in the list. Of course, this won't help the average user, so I'd argue that this property does not hold in practice.
- Very small list, giving only 8 bit entropy per word.
- Has not been translated yet

I think if we give up holding on that "phonetically distinct" property (as it does not really work out IMO), we can start discussing more secure (and localized) alternatives.

- The easiest step is to merge the even and the odd word list, giving one additional bit of entropy per word. This would quarter the chance of an attacker randomly guessing a code.
- Diceware uses a wordlist with 7,776 words, giving almost 13 bits of entropy. However, I haven't found any convincing translations yet.
- Linked from the python issue (below), I've found the [BIP 39](https://github.com/bitcoin/bips/blob/master/bip-0039/bip-0039-wordlists.md) word list (bitcoin enhancement proposal). It has several nice properties:
    - The first three or four characters uniquely identify a word, good for auto-completion
    - 12 bits of entropy (I think)
    - Translated into multiple languages. There's a list of criteria for each of the languages how the words were picked
- Don't use words for codes
    - Only 4 hex characters are needed for 16 bit entropy. That's still easy to remember: "C6A3"
    - Using the alphabet gives 36⁴ possibilities, or more than 20 bit entropy (we'd get the same with 5 hex digits FWIW)
    - Obvious problem is how to communicate casing (even if we say it doesn't matter)

If we have multiple word lists or code generation schemes, we then need a way for the user to choose one. CLI flags alone won't cut it, as specifying them every single time is tedious.

Python issue: https://github.com/magic-wormhole/magic-wormhole/issues/301

piegamesde:
**Update:**
The BIP word list translations sadly kind of died off. In any case, translating word lists is really hard and probably not worth it. I instead found https://www.eff.org/deeplinks/2016/07/new-wordlists-random-passphrases, which includes a really great one: 10.3 bits/word, three letter prefix uniqueness (the only one from the diceware category). If we stick to words, this would be it.

[...]

PR #194, opened 2023-04-28
Switch to BIP39 word list by default

lmfont:
Improves security slightly with default settings (16 bits -> 22 bits), and despite this the average word length is shorter (7.65 -> 5.40 chars) and (IMO) more easily remembered, communicated, and spelled.


afontenot:
I'm marking this as a draft because it's definitely unfinished (needs work on tests, etc), and also some feedback.

I realize BIP39 may not be the best theoretical wordlist approach, or anything like that, but IMO it's an easy change that's superior in every single way over the existing word list. It's worth moving to, even if it's ultimately temporary, for an easy win.

If you're interested in merging this, what would you like to do with the existing word list? Should it be left as dead code for some future word list switching feature? Removed?

piegamesde:
I'd like to point out https://github.com/magic-wormhole/magic-wormhole.rs/issues/106 and https://github.com/magic-wormhole/magic-wormhole-protocols/issues/31

afontenot:
Right, I had seen #106 before. I certainly see the appeal of three or four letter codes. My goal here was simply, while further changes are brewing, to go ahead and get an easy win while maintaining the current `number-word-word` format that users are used to.

PR #195, opened 2023-04-30
Combine connect methods

wuan:
## Description

This PR replaces the functions
```
    pub async fn connect_without_code(
        config: AppConfig<impl serde::Serialize + Send + Sync + 'static>,
        code_length: usize,
    ) -> Result<
        (
            WormholeWelcome,
            impl std::future::Future<Output = Result<Self, WormholeError>>,
        ),
        WormholeError,
    >
```
and
```
pub async fn connect_with_code(
        config: AppConfig<impl serde::Serialize + Send + Sync + 'static>,
        code: Code,
        expect_claimed_nameplate: bool,
    ) -> Result<(WormholeWelcome, Self), WormholeError> 
```
by separating the setup of the connection to the mailbox from the wormhole connection setup in two steps:

1. Create a `MailboxConnection` (with at least two variants):
    ```
    async MailboxConnection::create(
        config: AppConfig<V>,
        code_length: usize
    ) -> Result<Self, WormholeError>
    ```
    or

    ```
    async MailboxConnection::connect(
        config: AppConfig<V>,
        code: Code,
        allocate: bool,
    ) -> Result<Self, WormholeError>
    ```
3. Create the `Wormhole` connection using the `MailboxConnection`:
    ```
    async Wormhole::connect(
        mailbox_connection: MailboxConnection<impl serde::Serialize + Send + Sync + 'static>,
    ) -> Result<Self, WormholeError>
    ```

This has the advantage that setting up the initial connection and connecting to an existing mailbox have similar code. 

Waiting for the nameplate (within the generated code) from the initial connection has been made explicit. 

## Examples

The new `connect(...)` function can be used with or without code:

```
    let mailbox_connection = MailboxConnection::create(app_config, 2).await?;
    <show welcome message and show or transfer code>
    let wormhole = Wormhole::connect(mailbox_connection).await?;
```
or
```
    let mailbox_connection = MailboxConnection::connect(app_config, code, false).await?;
    <show welcome message>
    let wormhole = Wormhole::connect(mailbox_connection).await?;
```

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -63,6 +63,10 @@ impl From<std::convert::Infallible> for WormholeError {
  * The result of the client-server handshake
  */
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[deprecated(
+    since = "0.7.0",
+    note = "part of the response of `Wormhole::connect_without_code(...)` and `Wormhole::connect_with_code(...) please use 'MailboxConnection::create(...)`/`MailboxConnection::connect(..)` and `Wormhole::connect(mailbox_connection)' instead"
+)]
 pub struct WormholeWelcome {
     /** A welcome message from the server (think of "message of the day"). Should be displayed to the user if present. */
     pub welcome: Option<String>,
@@ -75,9 +79,8 @@ pub struct WormholeWelcome {
  * You can send and receive arbitrary messages in form of byte slices over it, using [`Wormhole::send`] and [`Wormhole::receive`].
  * Everything else (including encryption) will be handled for you.
  *
- * To create a wormhole, use the [`Wormhole::connect_without_code`], [`Wormhole::connect_with_code`] etc. methods, depending on
- * which values you have. Typically, the sender side connects without a code (which will create one), and the receiver side
- * has one (the user entered it, who got it from the sender).
+ * To create a wormhole, use the mailbox connection created via [`MailboxConnection::create`] or [`MailboxConnection::connect*`] with the [`Wormhole::connect`] method.
+ * Typically, the sender side connects without a code (which will create one), and the receiver side has one (the user entered it, who got it from the sender).
  *
  * # Clean shutdown
  *
@@ -87,6 +90,151 @@ pub struct WormholeWelcome {
  * Maybe a better way to handle application level protocols is to create a trait for them and then
  * to paramterize over them.
  */
+
+/// A `MailboxConnection` contains a `RendezvousServer` which is connected to the mailbox
+pub struct MailboxConnection<V: serde::Serialize + Send + Sync + 'static> {
+    /// A copy of `AppConfig`,
+    config: AppConfig<V>,
+    /// The `RendezvousServer` with an open mailbox connection
+    server: RendezvousServer,
+    /// The welcome message received from the mailbox server
+    pub welcome: Option<String>,
+    /// The mailbox id of the created mailbox
+    pub mailbox: Mailbox,
+    /// The Code which is required to connect to the mailbox.
+    pub code: Code,
+}
+
+impl<V: serde::Serialize + Send + Sync + 'static> MailboxConnection<V> {
+    /// Create a connection to a mailbox which is configured with a `Code` starting with the nameplate and by a given number of wordlist based random words.
+    ///
+    /// # Arguments
+    ///
+    /// * `config`: Application configuration
+    /// * `code_length`: number of words used for the password. The words are taken from the default wordlist.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # fn main() -> eyre::Result<()> { async_std::task::block_on(async {
+    /// use magic_wormhole::{transfer::APP_CONFIG, AppConfig, MailboxConnection};
+    /// let config = APP_CONFIG;
+    /// let mailbox_connection = MailboxConnection::create(config, 2).await?;
+    /// # Ok(()) })}
+    /// ```
+    pub async fn create(config: AppConfig<V>, code_length: usize) -> Result<Self, WormholeError> {
+        Self::create_with_password(
+            config,
+            &wordlist::default_wordlist(code_length).choose_words(),
+        )
+        .await
+    }
+
+    /// Create a connection to a mailbox which is configured with a `Code` containing the nameplate and the given password.
+    ///
+    /// # Arguments
+    ///
+    /// * `config`: Application configuration
+    /// * `password`: Free text password which will be appended to the nameplate number to form the `Code`
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # fn main() -> eyre::Result<()> { async_std::task::block_on(async {
+    /// use magic_wormhole::{transfer::APP_CONFIG, MailboxConnection};
+    /// let config = APP_CONFIG;
+    /// let mailbox_connection = MailboxConnection::create_with_password(config, "secret").await?;
+    /// # Ok(()) })}
+    /// ```
+    pub async fn create_with_password(
+        config: AppConfig<V>,
+        password: &str,
+    ) -> Result<Self, WormholeError> {
+        let (mut server, welcome) =
+            RendezvousServer::connect(&config.id, &config.rendezvous_url).await?;
+        let (nameplate, mailbox) = server.allocate_claim_open().await?;
+        let code = Code::new(&nameplate, &password);
+
+        Ok(MailboxConnection {
+            config,
+            server,
+            mailbox,
+            code,
+            welcome,
+        })
+    }
+
+    /// Create a connection to a mailbox defined by a `Code` which contains the `Nameplate` and the password to authorize the access.
+    ///
+    /// # Arguments
+    ///
+    /// * `config`: Application configuration
+    /// * `code`: The `Code` required to authorize to connect to an existing mailbox.
+    /// * `allocate`: `true`: Allocates a `Nameplate` if it does not exist.
+    ///               `false`: The call fails with a `WormholeError::UnclaimedNameplate` when the `Nameplate` does not exist.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # fn main() -> eyre::Result<()> { async_std::task::block_on(async {
+    /// use magic_wormhole::{transfer::APP_CONFIG, Code, MailboxConnection, Nameplate};
+    /// let config = APP_CONFIG;
+    /// let code = Code::new(&Nameplate::new("5"), "password");
+    /// let mailbox_connection = MailboxConnection::connect(config, code, false).await?;
+    /// # Ok(()) })}
+    /// ```
+    pub async fn connect(
+        config: AppConfig<V>,
+        code: Code,
+        allocate: bool,
+    ) -> Result<Self, WormholeError> {
+        let (mut server, welcome) =
+            RendezvousServer::connect(&config.id, &config.rendezvous_url).await?;
+        let nameplate = code.nameplate();
+        if !allocate {
+            let nameplates = server.list_nameplates().await?;
+            if !nameplates.contains(&nameplate) {
+                server.shutdown(Mood::Errory).await?;
+                return Err(WormholeError::UnclaimedNameplate(nameplate));
+            }
+        }
+        let mailbox = server.claim_open(nameplate).await?;
+
+        Ok(MailboxConnection {
+            config,
+            server,
+            mailbox,
+            code,
+            welcome,
+        })
+    }
+
+    /// Shut down the connection to the mailbox
+    ///
+    /// # Arguments
+    ///
+    /// * `mood`: `Mood` should give a hint of the reason of the shutdown
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # fn main() -> eyre::Result<()> { use magic_wormhole::WormholeError;
+    /// async_std::task::block_on(async {
+    /// use magic_wormhole::{transfer::APP_CONFIG, MailboxConnection, Mood};
+    /// let config = APP_CONFIG;
+    /// let mailbox_connection = MailboxConnection::create_with_password(config, "secret")
+    ///     .await?;
+    /// mailbox_connection.shutdown(Mood::Happy).await?;
+    /// # Ok(())})}
+    /// ```
+    pub async fn shutdown(self, mood: Mood) -> Result<(), WormholeError> {
+        self.server
+            .shutdown(mood)
+            .await
+            .map_err(WormholeError::ServerError)
+    }
+}
+
 #[derive(Debug)]
 pub struct Wormhole {
     server: RendezvousServer,
@@ -128,6 +276,11 @@ impl Wormhole {
      * do the rest of the client-client handshake and yield the [`Wormhole`] object
      * on success.
      */
+    #[deprecated(
+        since = "0.7.0",
+        note = "please use 'MailboxConnection::create(...) and Wormhole::connect(mailbox_connection)' instead"
+    )]
+    #[allow(deprecated)]
     pub async fn connect_without_code(
         config: AppConfig<impl serde::Serialize + Send + Sync + 'static>,
         code_length: usize,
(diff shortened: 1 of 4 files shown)


piegamesde:
Judging from the diff of the tests, I think this is a downgrade in API usability. Moreover, importing enum variants is frowned upon in many contexts (although I think it is kind of fine here?).

I think you can still make your deduplication by having the old API methods as shims delegating to the actual generic implementation.

Also keep in mind the use-case of connecting directly to a mailbox without nameplate, even though we don't currently make use of it. This is definitely a feature that will eventually be supported some time in the future.

wuan:
> Judging from the diff of the tests, I think this is a downgrade in API usability. Moreover, importing enum variants is frowned upon in many contexts (although I think it is kind of fine here?).

Having boolean values in APIs is some kind of anti-pattern. Enums help to group parameters for the different use cases and use names.

> I think you can still make your deduplication by having the old API methods as shims delegating to the actual generic implementation.

This is about deduplication and removing boolean flags from argument lists. 

> Also keep in mind the use-case of connecting directly to a mailbox without nameplate, even though we don't currently make use of it. This is definitely a feature that will eventually be supported some time in the future.

The change should not have removed any use-cases which have existed before. Can you give an example what will not work after the change?

piegamesde:
> Having boolean values in APIs is some kind of anti-pattern. Enums help to group parameters for the different use cases and use names for clarification.

I agree in general, but this was not my point. I was referring to having `WormholeConnect::GenerateCode` etc. in the file's global scope. But then if you don't import it like that the simple `WithCode` becomes `WormholeConnect::WithCode` or even `magic_wormhole::WormholeConnect::WithCode` …

> The change should not have removed any use-cases which have existed before. Can you give an example what will not work after the change?

I'm not saying it does, I'm just saying to keep it in mind so that it doesn't.

-----

Let's take a step back and explore some more API ideas that might our current issues with it.

- We could do a full-blown builder pattern
  - It would require two structs for the type state, and one of them would have a `expect_claimed_nameplate()` function or something like that
- We could have `WormholeConnect::connect(self, AppConfig)`
  - Downside is that the `Wormhole` struct has no constructor on itself anymore, but that's not uncommon in Rust APIs
  - Maybe call it `WormholeConnector` then

wuan:
> I agree in general, but this was not my point. I was referring to having `WormholeConnect::GenerateCode` etc. in the file's global scope. But then if you don't import it like that the simple `WithCode` becomes `WormholeConnect::WithCode` or even `magic_wormhole::WormholeConnect::WithCode` …

I do not see a problem here, as it still will be type safe and the user of the library can decide what to do adapting to the use case. 

And yes, a builder pattern can be added on top of this any time.

wuan:
> Hmm, I am sorry for being nitpicky but I'm still not convinced.
> 
> * The motivation of this PR is to remove the bool argument in the function signature, which it does and is a good thing.
> * The PR also claims to remove code duplication by merging two similar functions, but actually it has little effect. If you look at `initiate_connect`, almost all the relevant business logic is within the match/case distinction, except for a couple of lines of boilerplate code for input and return. But also this is not surprising, as all the duplicated code already got factored out into `connect_custom`.

The old code duplicates the function definition, calling `RendezvousServer::connect(...)` and building up the response. That is already duplication.

> * The `connect_with_code` and `connect_without_code` functions, while being similar, nevertheless have a distinct return type. This is now modeled using `connect` and `initiate_connect` functions. But which one of those should be used when? Previously the function signature just took care about that part, now users must think about it.

`initiate_connect(...)` is only there for the test setup. It should be removed from the public API to avoid confusion. Even worse, the slightly different return type of `connect_without_code(...)` has been added to a public API only for purpose of internal testing. That is something I would call unexpected and confusing.

> So overall this feels like a "one step forward one step back" kind of change to me. (Don't get me wrong, the proposed API is okay, but I don't see the value in replacing one okay API with small flaws by another API with other small flaws). I think this could be improved so that it does not have these problems, we could discuss that if desired. But as of now, IMHO this is mainly about the boolean argument flag, and for that there is a minimal solution:
> 
> You can give the argument a name by wrapping it in a tuple struct:
> 
> ```
> /// Wormhole.connect_with_code(config, code, Allocate(false))
> #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
> pub struct Allocate(pub bool);
> ```

I'm still not convinced and I propose to hide `initiate_connect(...)` from the user. In the end just having a single connect method which can be configured by a type safe parameter (Enums allow grouping of required parameters per use case) seem to be quite promising, especially as there are more ways to connect in the pipeline.

Besides that, I do not see the difference between importing a struct or an Enum.

> 
> The Gtk-rs bindings already make use of this pattern to map boolean arguments from the C side, and quite successfully so: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/signal/struct.Inhibit.html https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/source/struct.Continue.html
> 
> Also I chose `Allocate` as the name, because if we are touching the `expect_claimed_nameplate` flag while we are at it I'd like to invert it and call it "allocate", in accordance with the spec proposal. (It is an improvement Meejah came up with, and I really like it so why not also apply it here)

Renaming the boolean flag to "allocate" should be straightforward.

piegamesde:
> initiate_connect(...) is only there for the test setup. It should be removed from the public API to avoid confusion. Even worse, the slightly different return type of connect_without_code(...) has been added to a public API only for purpose of internal testing.

This is news to me, could you please elaborate? Unifying the return types / getting rid of `initiate_connect` would indeed be an interesting improvement.

wuan:
You are right, that is a general requirement. The basic problem seems to be that we want to access the generated code as a partial result, before we open the wormhole connection, which blocks until the other side enters exactly this code. That is for me the reason why we have that asymmetry in return types.

To overcome this, I now split up in creating the mailbox connection and creating the wormhole:

Setting up the wormhole on the code generating side looks like this:
```
let mailbox_connection = MailboxConnection::create(app_config, code_length).await?;
let code = mailbox_connection.code.clone();
<display/send code>
let wormhole = Wormhole::connect(mailbox_connection).await?;
```

and on the other side like this:

```
let mailbox_connection = MailboxConnection::connect(app_config, code, false).await?;
let wormhole = Wormhole::connect(mailbox_connection).await?;
```

which can even be simplified to, if you do not care about the welcome message:

```
let wormhole = Wormhole::connect_with_code(mailbox_connection, code, false).await?;
```

Does that help to make the steps more clear?

Names are still "in progress", better ideas are welcom.

src/core.rs
```diff
@@ -69,15 +69,75 @@ pub struct WormholeWelcome {
...
+
+    pub async fn connect(
+        config: AppConfig<V>,
+        code: Code,
+        allocate: bool,
+    ) -> Result<Self, WormholeError> {
+        let (mut server, welcome) =
+            RendezvousServer::connect(&config.id, &config.rendezvous_url).await?;
+        let nameplate = code.nameplate();
+        if allocate {
```
piegamesde, line 115:
```suggestion
        if !allocate {
```

The semantics have flipped with the rename

src/core.rs
```diff
@@ -211,14 +196,19 @@ impl Wormhole {
     ///
     /// If the [`RendezvousServer`] is not properly initialized, i.e. if the
     /// mailbox is not open.
-    pub async fn connect_custom(
-        mut server: RendezvousServer,
-        appid: AppID,
-        password: String,
-        app_versions: impl serde::Serialize + Send + Sync + 'static,
+    pub async fn connect(
+        mailbox_connection: MailboxConnection<impl serde::Serialize + Send + Sync + 'static>,
```
piegamesde, line 200:
```suggestion
        MailboxConnection {
            config,
            mut server,
            mailbox: _mailbox,
            code,
            welcome: _welcome,
        }: MailboxConnection<impl serde::Serialize + Send + Sync + 'static>,
```

I think nowadays one can destructure function arguments like this? Not sure if it is an improvement over doing it in a let bindings below. (Also, you can use `..` to ignore all remaining fields without having to list them)

src/core.rs
```diff
@@ -69,15 +69,75 @@ pub struct WormholeWelcome {
     pub code: Code,
 }
 
+pub struct MailboxConnection<V: serde::Serialize + Send + Sync + 'static> {
```
piegamesde, line 72: `MailboxServer`? `MailboxServerConnection`?

src/core.rs
```diff
@@ -69,15 +69,75 @@ pub struct WormholeWelcome {
     pub code: Code,
 }
```
piegamesde, line 70: This struct can be removed now I think

piegamesde (changes requested):
Looks interesting on the first look. Maybe always having a two-step connection even for the receive side is a good thing – with the old API, the receiver would only print the welcome message after the entire handshake had gone through. Now, it is more symmetric between both sides.

> which can even be simplified to the following, if you do not care about the welcome message:

I thought about proposing `mailbox_connection.connect_wormhole()` for that purpose, but then noticed that we have literally no use case for this as we always print the welcome message, and we should not encourage ignoring it

src/core.rs
wuan: Thanks, this should be fixed now and we have a pair of tests to verify that behaviour.

src/core.rs
wuan: Yes, did not now that. This works great, especially the `..`.

src/core.rs
wuan: So far I thought that the `MailboxServer` helps to create a `Mailbox` and here we have a connection to that `Mailbox`: `MailboxConnection`. I'm fine with both.

src/core.rs
wuan: It is still being used with the deprecated method(s). Just added a deprecated annotation here as well.

src/core.rs
wuan: As it would change the behaviour I will revert `connect_with_code(...)` to use `expect_claimed_nameplate` and mark it as deprecated. That should allow a smooth transition to the `allocate` argument.

src/core/test.rs
```diff
@@ -35,25 +40,67 @@ fn default_relay_hints() -> Vec<transit::RelayHint> {
     ]
 }
 
+#[async_std::test]
+pub async fn test_connect_with_unknown_code_and_allocate_passes() {
+    init_logger();
+
+    // the max nameplate number is 999, so this will not impact a real nameplate
```
piegamesde, line 47: We use a custom appid for tests, so there were no collisions in the first place

piegamesde (commented):
Code looking good so far. The doc comments still need adapting though

src/core/test.rs
wuan: Yes, that comment has already been there, can delete that.

src/core/test.rs
```diff
@@ -35,25 +40,67 @@ fn default_relay_hints() -> Vec<transit::RelayHint> {
...
+            assert!(false);
+        },
+    }
+}
+
 /** Send a file using the Rust implementation. This does not guarantee compatibility with Python! ;) */
 #[cfg(feature = "transfer")]
 #[async_std::test]
-pub async fn test_file_rust2rust() -> eyre::Result<()> {
+pub async fn test_file_rust2rust_deprecated() -> eyre::Result<()> {
```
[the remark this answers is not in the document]
piegamesde: Oops, should have been `#[allow(deprecated)]` instead; compiler is not happy at all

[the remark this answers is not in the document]
wuan: That should be fixed now.

wuan:
> Code looking good so far. The doc comments still need adapting though

After having lots of fun with doctests there should be some more doc comments now.

meejah:
Would being more like the Python API help here?

There, we do not conflate "connecting" or "creating" the Wormhole with the code at all -- instead you must call `.set_code()` or `.allocate_code()` (i.e. depending on whether you have one or wish to create one) at some point, before calling `.get_code()` (all code-paths call this).

This is more like "Builder" I suppose, superficially. But "the wormhole" has state, so there's no real getting around that.

You could also get rid of the bool in the API by declaring "the code" as a `Maybe`: if it's `Nothing` then allocate one. (I guess in Rust that's spelled `Option` and `None`).

wuan:
> You could also get rid of the bool in the API by declaring "the code" as a `Maybe`: if it's `Nothing` then allocate one. (I guess in Rust that's spelled `Option` and `None`).

That sounds different. In this case we always have a code. The boolean only determines the behaviour if the nameplate cannot be found: create it (`true`) or fail (`false`).

meejah:
How do you know if the nameplate can be found without trying to `claim` it?

piegamesde:
Yeah that is a temporary until we have the allocate flag in the `claim` command merged

meejah:
I don't anticipate merging that anytime soon; it breaks the protocol. But, okay, understood.

wuan:
> I don't anticipate merging that anytime soon; it breaks the protocol. But, okay, understood.

This PR only **renames** and inverts the semantics of the boolean parameter, it does not change functionality. We can even rename and invert the parameter to have no change at all.

meejah:
> This PR only renames and inverts the semantics [..]

I'm not talking about this PR in the sentence you quote.

wuan:
> I'm not talking about this PR in the sentence you quote.

Then I propose to add a reference to an issue / PR to link this discussion there.

src/core.rs
```diff
@@ -87,6 +90,162 @@ pub struct WormholeWelcome {
...
+    ///
+    /// A copy of `AppConfig`,
+    /// the `RendezvousServer` with the open mailbox connection,
+    /// the welcome message received from the mailbox server,
+    /// the mailbox id and
+    /// the Code which is required to connect to the mailbox.
+    ///
+    /// # Examples
+    ///
+    /// ```
```
piegamesde, line 123:
```suggestion
    /// ```no_run
```

Probably better to not run these tests, because unlike the "real" tests these use the production configuration (by design). Theoretically, you could silently set up a test mock so that the example still remains copypastable, but honestly that's probably not worth the effort.

src/core.rs
```diff
@@ -87,6 +90,162 @@ pub struct WormholeWelcome {
...
+    /// use futures::executor;
```
piegamesde, line 124: It's better to try and hide the setup stuff which is not directly linked to the usage. The example should be easy to copy-paste, and for that we may assume that the developer is already in an appropriate context (in this case, async and returning a `Result`).

You can have hidden setup lines by prefixing them with `#`, and the `async_std` crate does the following for their examples (I just adapted the Result type):

```suggestion
    /// # fn main() -> eyre::Result<()> { async_std::task::block_on(async {
```

(Full reference: https://docs.rs/async-std/latest/src/async_std/net/tcp/stream.rs.html#62-72)

piegamesde (changes requested):

src/core.rs
wuan: I already tried that, but it seemed to have no effect. Will add the `no_run` again.

src/core.rs
wuan: Fully agree here. Thanks for the hint. This works great.

src/core.rs
```diff
@@ -87,6 +90,162 @@ pub struct WormholeWelcome {
...
+
+    ///
+    /// Create a connection to a mailbox which is configured with a `Code` containing the nameplate and the given password.
+    ///
+    /// # Arguments
+    ///
+    /// * `config`: Application configuration
+    /// * `password`: Free text password which will be appended to the nameplate number to form the `Code`
+    ///
+    /// returns: Result<MailboxConnection<V>, WormholeError>
```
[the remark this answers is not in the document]
wuan: That line has been autogenerated by the IDE. Will drop all of them.

meejah:
> > I'm not talking about this PR in the sentence you quote.
> 
> Then I propose to add a reference to an issue / PR to link this discussion there.

I don't think it's relevant, but others can link it if they believe it is. I was just trying to clarify here that the PR piegames mentioned (and to which my reply pertained) wasn't _this_ one, as you seemed to believe with your reply.

The only relevance is that the "depend on `list` working" code in this repository is intended to be a temporary situation.

piegamesde (approved):
Looking good, only nits left. Cargo fmt is unhappy, but if you tell me that it runs fine on your machine then that's sufficient (we unfortunately have to use unstable for fmt so you may just have a different version with a different output).

Feel free to make the commit history tidy otherwise I'll just squash it into one commit. (I like "atomic" commits but arguably this entire PR is pretty atomic so it's fine)

wuan:
Thanks, rebased and squashed all commits.

PR #196, opened 2023-05-09
Initial set of changes for dilation

vu3rdd:
These commits add basic types etc for the manager state machine, makes dilation a command line option, refactors some non-dilation code a bit. The client can exchange `please` message with the python dilation client and is able to receive `connection-hints` message.

The plan is to send PRs so that we minimize the size of the subsequent PRs and make it ready for review and merge them as the development progresses rather than wait till the end and send a huge PR that would be hard to review.

Keep in mind that we (@ewanas, @vu3rdd, @wuan) are rust newbies, so expect parts of the code to be non-idiomatic.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ base64 = "0.20.0"
 futures_ringbuf = "0.3.1"
 time = { version = "0.3.7", features = ["formatting"] }
 instant = { version = "0.1.12", features = ["wasm-bindgen"] }
+mockall_double = "0.3.0"
 
 derive_more = { version = "0.99.0", default-features = false, features = ["display", "deref", "from"] }
 thiserror = "1.0.24"
@@ -81,7 +82,7 @@ getrandom_2 = { package = "getrandom", version = "0.1.16", features = ["js-sys"]
 [dev-dependencies]
 env_logger = "0.10.0"
 eyre = "0.6.5"
-mockall = "0.11"
+mockall = "0.11.4"
 
 [features]
 transit = ["socket2", "stun_codec", "if-addrs", "bytecodec", "async-trait", "noise-protocol", "noise-rust-crypto"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,8 +85,9 @@ eyre = "0.6.5"
 [features]
 transit = ["socket2", "stun_codec", "if-addrs", "bytecodec", "async-trait", "noise-protocol", "noise-rust-crypto"]
 transfer = ["transit", "async-tar", "rmp-serde"]
+dilation = ["transfer"]
 forwarding = ["transit", "rmp-serde"]
-default = ["transit", "transfer"]
+default = ["transfer", "dilation"]
 all = ["default", "forwarding"]
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,7 @@ getrandom_2 = { package = "getrandom", version = "0.1.16", features = ["js-sys"]
 [dev-dependencies]
 env_logger = "0.10.0"
 eyre = "0.6.5"
+mockall = "0.11"
 
 [features]
 transit = ["socket2", "stun_codec", "if-addrs", "bytecodec", "async-trait", "noise-protocol", "noise-rust-crypto"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,9 +31,10 @@ log = "0.4.13"
 # zeroize = { version = "1.2.0", features = ["zeroize_derive"] }
 base64 = "0.20.0"
 futures_ringbuf = "0.3.1"
+async-trait = "0.1.57"
+mockall_double = "0.3.0"
 time = { version = "0.3.7", features = ["formatting"] }
 instant = { version = "0.1.12", features = ["wasm-bindgen"] }
-mockall_double = "0.3.0"
 
 derive_more = { version = "0.99.0", default-features = false, features = ["display", "deref", "from"] }
 thiserror = "1.0.24"
@@ -46,7 +47,6 @@ percent-encoding = { version = "2.1.0" }
 
 stun_codec = { version = "0.2.0", optional = true }
 bytecodec = { version = "0.4.15", optional = true }
-async-trait = { version = "0.1.57", optional = true }
 noise-protocol = { version = "0.1.4", optional = true }
 noise-rust-crypto = { version = "0.5.0", optional = true }
 
@@ -85,9 +85,9 @@ eyre = "0.6.5"
 mockall = "0.11.4"
 
 [features]
-transit = ["socket2", "stun_codec", "if-addrs", "bytecodec", "async-trait", "noise-protocol", "noise-rust-crypto"]
+transit = ["socket2", "stun_codec", "if-addrs", "bytecodec", "noise-protocol", "noise-rust-crypto"]
 transfer = ["transit", "async-tar", "rmp-serde"]
-dilation = ["transfer"]
+dilation = ["transit"]
 forwarding = ["transit", "rmp-serde"]
 default = ["transfer", "dilation"]
 all = ["default", "forwarding"]
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -17,8 +17,9 @@ use std::{
     io::Write,
     path::{Path, PathBuf},
 };
+use serde_json::Value;
 
-use magic_wormhole::{forwarding, transfer, transit, MailboxConnection, Wormhole};
+use magic_wormhole::{forwarding, transfer, transit, dilation, MailboxConnection, Wormhole};
 
 fn install_ctrlc_handler(
 ) -> eyre::Result<impl Fn() -> futures::future::BoxFuture<'static, ()> + Clone> {
@@ -335,7 +336,7 @@ async fn main() -> eyre::Result<()> {
                     code,
                     Some(code_length),
                     true,
-                    transfer::APP_CONFIG,
+                    dilation::APP_CONFIG_SEND,
                     Some(&sender_print_code),
                     clipboard.as_mut(),
                 )),
@@ -373,7 +374,7 @@ async fn main() -> eyre::Result<()> {
                     code,
                     Some(code_length),
                     true,
-                    transfer::APP_CONFIG,
+                    dilation::APP_CONFIG_SEND,
                     Some(&sender_print_code),
                     clipboard.as_mut(),
                 ));
@@ -412,14 +413,14 @@ async fn main() -> eyre::Result<()> {
             ..
         } => {
             let transit_abilities = parse_transit_args(&common);
-            let (wormhole, _code, relay_hints) = {
+            let (mut wormhole, _code, relay_hints) = {
                 let connect_fut = Box::pin(parse_and_connect(
                     &mut term,
                     common,
                     code,
                     None,
                     false,
-                    transfer::APP_CONFIG,
+                    dilation::APP_CONFIG_RECEIVE,
                     None,
                     clipboard.as_mut(),
                 ));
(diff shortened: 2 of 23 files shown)


Cargo.toml
```diff
@@ -81,12 +82,14 @@ getrandom_2 = { package = "getrandom", version = "0.1.16", features = ["js-sys"]
 [dev-dependencies]
 env_logger = "0.10.0"
 eyre = "0.6.5"
+mockall = "0.11.4"
 
 [features]
 transit = ["socket2", "stun_codec", "if-addrs", "bytecodec", "async-trait", "noise-protocol", "noise-rust-crypto"]
 transfer = ["transit", "async-tar", "rmp-serde"]
+dilation = ["transfer"]
```
piegamesde, line 90: This looks very much the wrong way around: Dilation should not depend on anything except the core protocol.

cli/src/main.rs
```diff
@@ -253,6 +256,9 @@ struct WormholeCli {
     /// Enable logging to stdout, for debugging purposes
     #[clap(short = 'v', long = "verbose", alias = "log", global = true)]
     log: bool,
+    /// Enable dilation
+    #[clap(long = "with-dilation", alias = "with-dilation", global = true)]
+    with_dilation: bool,
```
piegamesde, line 261: This should not be a global flag, as not all commands may make use of it. It would best fit into `CommonSenderArgs` and `CommonReceiverArgs`, but `CommonArgs` would be fine too for now. (If you want to be *very correct*, that flag would belong to a new `CommonTransferArgs` or something)

cli/src/main.rs
```diff
@@ -486,7 +498,7 @@ async fn main() -> eyre::Result<()> {
                 })
                 .collect::<Result<Vec<_>, _>>()?;
             loop {
-                let mut app_config = forwarding::APP_CONFIG;
+                let mut app_config = forwarding::APP_CONFIG.with_dilation(app.with_dilation);
```
piegamesde, line 501: Port forwarding does not make use of Dilation

src/core.rs
```diff
@@ -353,6 +360,9 @@ impl Wormhole {
         /* Send versions message */
         let mut versions = key::VersionsMessage::new();
         versions.set_app_versions(serde_json::to_value(&config.app_version).unwrap());
+        if config.with_dilation {
+            versions.enable_dilation();
```
piegamesde, line 364: Does this compile with Dilation disabled? The CI has a test suite which tests that various feature flag combinations compile, Dilation should definitely be added to that list

src/core.rs
```diff
@@ -448,17 +471,17 @@ impl Wormhole {
      * used by upper layer protocols. We distinguish between the different layers
      * on which a serialization error happened, hence the double `Result`.
      */
-    pub async fn receive_json<T>(&mut self) -> Result<Result<T, serde_json::Error>, WormholeError>
+    pub async fn receive_json<T>(&mut self) -> Result<T, WormholeError>
```
piegamesde, line 474: Please undo this change again, see the docstring

src/core.rs
```diff
@@ -563,17 +615,25 @@ impl From<String> for AppID {
...
+    derive_more::Display,
+    derive_more::Deref,
 )]
 #[serde(transparent)]
 #[display(fmt = "MySide({})", "&*_0")]
 pub struct MySide(EitherSide);
 
 impl MySide {
-    pub fn generate() -> MySide {
+    pub fn generate(n: usize) -> MySide {
```
piegamesde, line 633:
```suggestion
    pub fn generate(length: usize) -> MySide {
```

src/core/key.rs
```diff
@@ -102,24 +103,40 @@ pub fn make_pake(password: &str, appid: &AppID) -> (Spake2<Ed25519Group>, Vec<u8
     (pake_state, pake_msg_ser)
 }
 
-#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
 pub struct VersionsMessage {
-    #[serde(default)]
-    pub abilities: Vec<String>,
```
piegamesde, line 108: The `abilities` field is still part of the protocol I think

src/dilation.rs
```diff
@@ -0,0 +1,395 @@
...
+#[double]
+use crate::dilation::manager::ManagerMachine;
+
+use super::AppID;
+
+mod api;
+mod events;
+mod manager;
+
+const APPID_RAW: &str = "lothar.com/wormhole/text-or-file-xfer";
```
piegamesde, line 29: This looks wrong to have here in the dilation module, as well as the other app config and transfer related stuff

src/dilation.rs
```diff
@@ -0,0 +1,395 @@
...
+        self.wormhole
+            .borrow_mut()
+            .send_json_with_phase(command, Phase::dilation)
+            .await
+    }
+}
+
+pub struct DilatedWormhole {
+    wormhole: WormholeConnection,
+    side: MySide,
```
piegamesde, line 142: We already distinguish between `MySide` and `OtherSide` in the type system. Given that the Wormhole itself has a side too, maybe add new dilation specific "side" types to prevent confusing them?

src/dilation.rs
```diff
@@ -0,0 +1,395 @@
+use std::{
```
piegamesde, line 1: I've wanted to move `src/transit.rs` to `src/transit/mod.rs` (and same for the others) for a while now, so preferably put this into `src/dilation/mod.rs` from the start

src/dilation/events.rs
```diff
@@ -0,0 +1,84 @@
...
+    },
+    RxReconnect,
+    RxReconnecting,
+    ConnectionMade,
+    ConnectionLostLeader,
+    ConnectionLostFollower,
+    Stop,
+}
+
+impl std::fmt::Display for ManagerEvent {
```
piegamesde, line 53: This would be easier to do with the derive macro

piegamesde (commented):
Did a rather quick first pass, still need to look at the actual Dilation code some more.

cli/src/main.rs
wuan: Should be fixed now -> We have a `CommonTransferArgs` enum.

cli/src/main.rs
wuan: Fixed as well

src/core.rs
wuan: We now have a
```
#[cfg(feature = "dilation")]
```
in front of this if.

src/core.rs
wuan: Done

src/core.rs
```diff
@@ -485,9 +508,32 @@ impl Wormhole {
     }
 }
 
+/**
+ * create a dilated wormhole
+ */
+#[cfg(feature = "dilation")]
+pub fn dilate(wormhole: Wormhole) -> Result<DilatedWormhole, WormholeError> {
```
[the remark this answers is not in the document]
wuan: Can give it a try. The wormhole object needs to be moved into `DilatedWormhole`.

src/core.rs
wuan: Thanks, this should be updated.

src/core/key.rs
wuan: That should be restored now.

src/dilation.rs
wuan: We currently have `transfer::APP_ID` with `transfer::APP_CONFIG` and also the same values in `forwarding::`. What do you propose we should do here?

src/dilation.rs
wuan: `src/dilation.rs` is now moved to `src/dilation/mod.rs`

src/dilation/events.rs
wuan: Fixed as well now.

src/dilation.rs
piegamesde: Um I very much hope that `transfer::APP_ID` is not the same as `forwarding::APP_ID` ^^

wuan: So should we go with something like `lothar.com/wormhole/dilation` then? Anyway that should be easy to change later in the project.

Cargo.toml
wuan: Should be fixed now.

src/dilation.rs
meejah: I _believe_ we can use the same AppId for both classic and next-gen transfer.

The main consideration here is interop: do we want applications to have the _possibility_ of inter-operating with classic clients? I believe it's possible and if so would prefer that behavior for the CLI when possible.

In any case, I think @piegamesde point (and the reason I agree) is that the AppId isn't up to Dilation at all but to the "application protocol" (which in this case will be "Dilated File Transfer" for next-gen transfer). So, there shouldn't be any app-id in the Dilation part (it should be a parameter given from the application-level protocol / user of Dilation).

More generally, there shouldn't be any "file-transfer specific" stuff in the Dilation layer at all.

wuan: We are now using `transfer::APP_ID`. The setup might still require some cleanup.

wuan: And we still use `transit::{DirectHint, Hints, RelayHint},` which is part of dilation as we have a hint exchange here as well.

piegamesde: I'd like to see as much code reuse between transit and dilation as possible. I don't think it is strictly necessary to move shared bits out of the `transit` module; it is fine for me if Dilation just imports these bits

wuan: That sound good, I added inclusion of transit when dilation is selected to fix that now.

wuan:
These are some questions raised by @piegamesde which we talked about and the solutions we agreed on:

The core module knows as little as possible about Dilation, because Dilation generally is opt-in (only thing it needs to know is to tell the other side "no I don't Dilate" if necessary, but even that should be a no-op because of backwards compatibility …)

* Finding: Core module contains VersionsMessage which has properties `can_dilate` and `dilation_abilities`.
  * Ignore for now
* Finding: Wormhole object has method `dilate()`
  * Encapsulate within feature flag

Dilation knows nothing about file transfer, because it may be used as the basis for any application protocol

* Finding We have an enum `FileTransferV2Mode` in `dilation/mod.rs` (Defining modes of file transfer application), `AppVersions` (in `dilation/mod.rs`) references `DilatedTransfer` (references `FileTransferV2Mode`)
  * Create a dilated file transfer module for these. It should be a module outside of `dilation` so that we only use the public API of dilation. This can be merged with the upcoming file transfer v2 module later.

Applications who use file transfer do not have to interact with Dilation. They just say "I want v2" and stuff just happens for them
* As of now, we switch to dilation if both sides can handle it, but only for the receive action.
* There is already a prototype created by piegames
* For now
  * Dilated file transfer module should assume that dilation is available. Negotiation part should be done elsewhere.
  * In the end there should be a file transfer which uses v2 (with dilation) if possible and fall back to v1.

meejah:
> Dilation generally is opt-in (only thing it needs to know is to tell the other side "no I don't Dilate" if necessary

There's no reason to say "I don't do Dilation": if a peer doesn't indicate support for something, it doesn't support it.

> Finding: Wormhole object has method dilate()    Encapsulate within feature flag

Applications will still want to opt-in to Dilation at runtime.

That is, plenty of "wormhole-using" applications may have no use or need for Dilation (e.g. magic-folder invites never use a bulk-data connection) so it seems you'd still want both things to me (a feature-flag, but also a `dilate()` or similar method for the application to signal "I actually want to Dilate this wormhole").

piegamesde:
> There's no reason to say "I don't do Dilation": if a peer doesn't indicate support for something, it doesn't support it.

You're totally right, that was just bad wording.

> Applications will still want to opt-in to Dilation at runtime.

This is covered by enabling the dilation feature flag and then (not) calling the `dilate()` method. Without the dilation feature, that method simply won't exist then.

src/core.rs
```diff
@@ -501,8 +497,10 @@ pub enum Mood {
     Unwelcome,
 }
 
+pub const APPID_RAW: &str = "lothar.com/wormhole/text-or-file-xfer";
```
piegamesde, line 500: There are some compiler warnings left, like this one

src/core.rs
```diff
@@ -563,17 +565,25 @@ impl From<String> for AppID {
 
 // MySide is used for the String that we send in all our outbound messages
 #[derive(
-    PartialEq, Eq, Clone, Debug, Deserialize, Serialize, derive_more::Display, derive_more::Deref,
+    PartialOrd,
```
piegamesde, line 568: See https://github.com/magic-wormhole/magic-wormhole.rs/pull/196#discussion_r1191112305

src/core.rs
```diff
@@ -589,7 +599,15 @@ impl MySide {
 
 // TheirSide is used for the string that arrives inside inbound messages
 #[derive(
-    PartialEq, Eq, Clone, Debug, Deserialize, Serialize, derive_more::Display, derive_more::Deref,
+    PartialOrd,
```
piegamesde, line 602: Same here

src/core/key.rs
```diff
@@ -133,6 +152,7 @@ pub fn build_version_msg(
     let phase = Phase::VERSION;
     let data_key = derive_phase_key(side, key, &phase);
     let plaintext = serde_json::to_vec(versions).unwrap();
+    println!("versions message before encryption: {:?}", plaintext);
```
piegamesde, line 155: Use `log::debug` instead

piegamesde:
The CI failure may be nightly, but since there has been a new stable release since this was committed it might fail there too.

A quick dig at dependencies' issues tells me that updating the dependencies / `cargo update` should fix the CI failure.

wuan:
Force pushed the compacted branch. The original branch is still available: https://github.com/LeastAuthority/magic-wormhole.rs/tree/dilation-1-original

PR #202, opened 2023-07-06
Feat/completion

### LINKED ISSUE #147: Tab completion
When asking for a code input, pressing tab should complete the word that is currently typed.  At the moment, we are using the `dialoguer` crate for input. I don't know if they provide support for tab completion. Otherwise, I'd be fine with switching to a different implementation (different crate or a custom helper function).

a-moreira:
closes #147


a-moreira:
this implementation is not ideal because the trait `Completion` in `dialoguer` expects only one completion (a `Option<String>`, while in our cause it would be better to expect a vector with possible completions. I'm open to other approaches, maybe using the crate `inquire` instead of `dialoguer` in the repo, which also provides Autocompletion. What do you think? @piegamesde

a-moreira:
btw also ran `clippy` in other parts of the repo, hope it's not a problem.

a-moreira:
some tests still breaking, will fix soon.

@piegamesde maybe it's not worth it to change to `dialoguer`, then we would need a way to expose the API you had written in a "loop" to get the input.

piegamesde:
I think dialoguer is fine for now. It only does completion and not suggestions, meaning that if there are multiple matching values no completion should be performed. This is not optimal, but okay for now (the Python implementation does it like this as well).

However, the magic wormhole library should not depend on dialoguer, that dependency should be exclusive to the CLI. Therefore you need to move the respective code out.

a-moreira:
@piegamesde I'm getting a cyclic dependency error when trying to import the CLI code into the root crate to get the `wordlist` mod, which is necessary in `src/core.rs`. Do we need to have the CLI as a separate crate in the root? Maybe we can change the project structure a little bit to have the CLI as a `mod` inside the crate root or another approach.

piegamesde:
CLI is separate from the rest on purpose, and if you are getting a cyclic dependency error that indicates you are doing something wrong. The library crate should never depend on the CLI or CLI code, and you trying to import CLI code in the library is what causes your error.

a-moreira:
@piegamesde you are right. I used an approach of wrapping the main wordlist struct in a new type so I can implement the `dialoguer` traits inside the CLI, thus splitting the code for each crate according to its necessary functionalities. I can change that if you think it's not good. Also removed a couple of tests that don't make sense anymore now that it doesn't return a vector of suggestions.

a-moreira:
one `cargo test` action seems to be failing, but the others not. Locally it runs ok.

src/core/wordlist.rs
```diff
@@ -1,63 +1,19 @@
...
-}
-
-impl Wordlist {
-    #[cfg(test)]
-    pub fn new(num_words: usize, words: Vec<Vec<String>>) -> Wordlist {
-        Wordlist { num_words, words }
-    }
-
-    #[allow(dead_code)] // TODO make this API public one day
-    pub fn get_completions(&self, prefix: &str) -> Vec<String> {
```
piegamesde, line 23: This method should be kept, and the wrapper for Dialoguer in the CLI should make use of it.

src/core/wordlist.rs
```diff
@@ -97,13 +53,25 @@ fn load_pgpwords() -> Vec<Vec<String>> {
...
-pub fn default_wordlist(num_words: usize) -> Wordlist {
-    Wordlist {
+pub fn default_wordlist(num_words: usize) -> PgpWordList {
+    PgpWordList {
         num_words,
         words: load_pgpwords(),
     }
 }
 
+pub fn vecstrings(all: &str) -> Vec<String> {
```
piegamesde, line 63: This function is really weird. At the very least it should have a descriptive name and documentation to tell what it does, but honestly I think it probably shouldn't exist / not be public in the first place.

cli/src/main.rs
```diff
@@ -1172,6 +1178,74 @@ async fn receive_inner_v2(
...
 }
 
+use dialoguer::Completion;
+use magic_wormhole::core::wordlist;
+use std::{collections::HashMap, fs};
+
+struct WordList(PgpWordList);
+
+impl Default for WordList {
+    fn default() -> Self {
```
piegamesde, line 1188: This looks like a re-implementation of `load_pgpwords`. Why not make use of that?

cli/src/main.rs
```diff
@@ -1172,6 +1178,74 @@ async fn receive_inner_v2(
...
+                    suffix.push('-');
+                }
+
+                completions.push(suffix);
+            }
+        }
+        if completions.len() == 1 {
+            Some(completions.first().unwrap().clone())
+        } else {
+            // TODO: show vector of suggestions somehow
```
piegamesde, line 1242: I'd consider that out of scope for now.

piegamesde (changes requested):

Issue #145, opened 2022-03-18
Add a man page


piegamesde:

werdahias:
fwiw, clap-mangen generates a man page a build time if you use clap for argument parsing. I can offer to write a "classical" man page and make a PR.

PR #226, opened 2024-05-16
Added manpage for magic-wormhole

werdahias:
Closes: #145


piegamesde:
I think I'd prefer a generated man page, as that should be better for long-term maintenance. Without at least some tooling yelling at you, it's too easy to modify the CLI options and then forget about the man page.

werdahias:
Sure, do you need me to push a relicensed version?

PR #228, opened 2024-06-13
ci(macos): added to releases

FrancescElies:


FrancescElies:
There is a windows test failing but that doesn't have to do with this pr.

PR #247, opened 2024-07-26
Fix dual-stack socket usage on Windows

himikof:


src/transit/transport.rs
```diff
@@ -72,6 +72,7 @@ impl<T> TransitTransport for T where T: AsyncRead + AsyncWrite + std::any::Any +
 #[cfg(not(target_family = "wasm"))]
 pub(super) fn set_socket_opts(socket: &socket2::Socket) -> std::io::Result<()> {
     socket.set_nonblocking(true)?;
+    socket.set_only_v6(false)?;
```
piegamesde, line 75:
```suggestion
    /* Explicitly make the socket dual-stack, otherwise IPv4 won't work on some platforms where 
     * it is enabled by default (e.g. Windows)
     */
    socket.set_only_v6(false)?;
```

himikof: Thanks, I've added a slightly expanded version of this comment to the commit.

PR #325, opened 2025-12-10
Add text message sending support

yabobay:


felinira:
I am deeply unhappy with the entire `Offer` API in general. However refactoring the whole thing to an enum would probably make it even more convoluted than before.

Part of why the offer API is how it is is because WASM is rather strange to support. But we can probably still do much better.

I am not sure I have an easy solution here that makes Offer less complicated instead of more. (And I certainly wouldn't expect from you to refactor the entire thing)

yabobay:
So.... what's going on? Idk if bumping is considered rude but bump

PR #351, opened 2026-04-12
cargo: replace macro_rules_attribute

officialasishkumar:
The RustSec report for `paste` is still applicable because `macro_rules_attribute` pulls `paste` into the dev-dependency graph. The project only uses `macro_rules_attribute` to apply a local async-test wrapper macro, so this switches those tests to `macro_rules_attr`, which supports the same `#[apply(test)]` use case without depending on `paste`.

Changes:
- Replace the workspace `macro_rules_attribute` dependency with `macro_rules_attr`.
- Update the affected async tests in `src/core/test.rs` from `macro_rules_attribute::apply(test!)` to `#[apply(test)]`.
- Update `Cargo.lock` so `paste` and `macro_rules_attribute` are removed.
- Add a changelog entry under Unreleased.

Validation:
- `cargo metadata --locked --format-version=1 --no-deps`
- `rg -n "paste|macro_rules_attribute" Cargo.lock Cargo.toml src/core/test.rs` returned no matches
- `cargo tree --locked --workspace --all-features --all-targets | rg "macro_rules_attr|macro_rules_attribute|paste"` shows only `macro_rules_attr`
- `cargo fmt --check`
- `cargo check --workspace --all-features --locked`
- `cargo test --workspace --all-features --no-run --locked`
- `git diff --check`

Fixes #340

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- dev: Replace the `macro_rules_attribute` test helper to remove the unmaintained `paste` dependency
+
 ## [0.8.0] - 2026-03-26
 
 ### Added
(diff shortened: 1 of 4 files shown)


felinira:
thanks!

PR #381, opened 2026-07-04
Make the STUN server configurable (CLI flag + API parameter)

### LINKED ISSUE #377: make STUN server configurable
https://github.com/magic-wormhole/magic-wormhole.rs/blob/e5c4cba668b54119b08314e271ae1fa7329bc6a2/src/transit.rs#L46

I think it's feasible to add an api/command-line optinon to configure this, in case users (like me) want to use their own STUN server

akagifreeez:
Closes #377

## What & why

Adds a `stun_server` parameter threaded from the CLI down to `transit::init()`, mirroring the existing relay-server pattern:

- `--stun-server <HOSTNAME:PORT>` CLI flag + `WORMHOLE_STUN_SERVER` env var (symmetric with `--relay-server` / `WORMHOLE_RELAY_URL`)
- `PUBLIC_STUN_SERVER` renamed to `pub const DEFAULT_STUN_SERVER` (symmetric with `DEFAULT_RELAY_SERVER`)
- `transit::init()` takes `stun_server: Option<&str>`, falling back to `DEFAULT_STUN_SERVER` when `None`; public `transfer`/`forwarding` entry points take `Option<String>` (owned, since the CLI spawns them as `'static` tasks) and pass it down via `.as_deref()`

The default (`stun.piegames.de:3478`) is **unchanged** — per @felinira's comment on the issue, picking an alternative server is a separate concern; this PR just makes it configurable so users can switch without a fork.

If the configured value fails to resolve, behavior matches the existing STUN failure path: a warning is logged and the connection falls back to a plain socket.

## Breaking change

Nine public library functions (`transit::init`, `transfer::{send,request,request_file,send_file,send_folder,send_file_or_folder}`, `forwarding::{serve,connect}`) gain a required `stun_server` argument. Recorded in the CHANGELOG under `### Changed` with the `[lib][breaking]` convention.

## How to test

- `wormhole-rs send --help` shows `--stun-server <HOSTNAME:PORT>` and `[env: WORMHOLE_STUN_SERVER=]`
- `cargo test --workspace --features all,native-tls,experimental` — all green (includes 2 new clap parsing tests for the flag/default)
- Ran the same checks as CI locally: `cargo fmt --check`, `cargo clippy --all-features`, per-feature builds (`transit`/`transfer`/`forwarding`/no-default), `magic-wormhole-cli --features=all`, and `wasm32-unknown-unknown` builds — all passing

## Notes

This PR was implemented with LLM assistance; all changes were human-reviewed and every check above was run locally before submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - lib: support for encrypted websocket connections through `futures-rustls` as a future replacement for the `async-tls` dependency
+- lib, cli: allow configuring a custom STUN server via `transit::init`'s new `stun_server` parameter, or the CLI's `--stun-server` flag / `WORMHOLE_STUN_SERVER` environment variable
+
+### Changed
+
+- \[lib\]\[breaking\] changed the signature of `transit::init`, `transfer::send`, `transfer::request`, `transfer::request_file`, `transfer::send_file`, `transfer::send_folder`, `transfer::send_file_or_folder`, `forwarding::serve` and `forwarding::connect` to require a new `stun_server: Option<&str>` argument
 
 ## [0.8.1] - 2026-05-07
 
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -126,6 +126,9 @@ struct CommonArgs {
         env = "WORMHOLE_RELAY_URL",
     )]
     relay_server: Vec<url::Url>,
+    /// Use a custom STUN server for NAT traversal (direct connection discovery).
+    #[arg(long, value_name = "HOSTNAME:PORT", env = "WORMHOLE_STUN_SERVER")]
+    stun_server: Option<String>,
     /// Use a custom rendezvous server. Both sides need to use the same value in order to find each other.
     #[arg(long, value_name = "ws://example.org", value_hint = clap::ValueHint::Url, env = "WORMHOLE_MAILBOX_URL")]
     rendezvous_server: Option<url::Url>,
@@ -330,7 +333,7 @@ async fn async_main() -> eyre::Result<()> {
             let offer = make_send_offer(files, file_name).await?;
 
             let transit_abilities = parse_transit_args(&common);
-            let (wormhole, _code, relay_hints) = match util::cancellable(
+            let (wormhole, _code, relay_hints, stun_server) = match util::cancellable(
                 Box::pin(parse_and_connect(
                     &mut term,
                     common,
@@ -349,7 +352,14 @@ async fn async_main() -> eyre::Result<()> {
                 Err(_) => return Ok(()),
             };
 
-            Box::pin(send(wormhole, relay_hints, offer, transit_abilities)).await?;
+            Box::pin(send(
+                wormhole,
+                relay_hints,
+                stun_server,
+                offer,
+                transit_abilities,
+            ))
+            .await?;
         },
         WormholeCommand::SendMany {
             tries,
@@ -365,7 +375,7 @@ async fn async_main() -> eyre::Result<()> {
             ..
         } => {
             let transit_abilities = parse_transit_args(&common);
-            let (wormhole, code, relay_hints) = {
+            let (wormhole, code, relay_hints, stun_server) = {
                 let connect_fut = Box::pin(parse_and_connect(
                     &mut term,
                     common,
@@ -385,6 +395,7 @@ async fn async_main() -> eyre::Result<()> {
 
             Box::pin(send_many(
                 relay_hints,
+                stun_server,
                 &code,
                 files,
                 file_name,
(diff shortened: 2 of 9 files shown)


felinira:
The idea is good. But this clutters 9 public API functions with the same argument, which is quite a lot and these methods are already very cluttered. Can't this just be stored in `Wormhole` struct with a setter so that it only has to be set once?

akagifreeez:
Thanks for taking a look!

I agree the repeated argument isn't nice. Before I rework it, two things I'd like to settle, since they change the shape of the fix:

**`transit::init` has no `Wormhole`.** Of the 9 functions, 8 (`transfer::send`/`request`/`request_file`/`send_file`/`send_folder`/`send_file_or_folder` and `forwarding::serve`/`connect`) take `wormhole: Wormhole` as their first argument, so a setter works cleanly for those. `transit::init` is the exception — it would have to keep taking the STUN server explicitly. Is that split acceptable to you?

**`relay_hints` is threaded through exactly the same 9 functions**, and it's the same kind of setting. If STUN moves onto `Wormhole` and relay hints stay as an argument, the two end up being configured in different places. So which would you prefer:

1. Move only `stun_server` onto `Wormhole` — smallest diff, but asymmetric with `relay_hints`; or
2. Move both — either two setters on `Wormhole`, or a single transit-config struct that `Wormhole` holds and `transit::init` also accepts. That one actually removes the clutter you're describing, at the cost of a bigger breaking change.

Either is fine by me, and I'd rather match the API you want than guess. If you also have a preference on the setter shape (`set_stun_server(&mut self, …)` vs a builder-style `with_stun_server()`), let me know and I'll rework the PR accordingly.

felinira:
Interesting questions which makes me rethink this completely. The approach I recommended before with just storing it in Wormhole is kinda conflating two things here on the API level:

a) The configuration for the transit connection, relay hints, stun server etc.
b) The configuration for the wormhole, which fundamentally doesn't know anything about STUN or relay servers, and many uses of Wormhole don't actually do the transit protocol

This makes me rethink the approach entirely. Storing it in Wormhole isn't the best idea either. So basically the API for file sending works like this:

`wormhole::MailboxConnection::create`
`wormhole::Wormhole::connect`
`wormhole::transfer::send_file`

So we don't *have* a nice place in this API to store configuration methods. I guess we can make it though.

Okay, so if you're up to it my idea is now the following:

1. Introduce a `transit::TransitConfig` struct:

```rust
struct TransitConfig {
    abilities: Abilities,
    relay_hints: Vec<RelayHint>,
    stun_server: Option<url::Url>
}
```

and this API

```rust
impl TransitConfig {
pub fn new(abilities: Abilities) -> Self
pub fn abilities(&self) -> Abilities
pub fn set_abilities(&mut self, abilities: Abilities)
pub fn relay_hints(&self) -> Vec<RelayHint>
pub fn set_relay_hints(&mut self, relay_hints: Vec<RelayHint>)
pub fn stun_server(&self) -> Option<url::Url>
pub fn set_stun_server(&mut self, stun_server: Option<url::Url>)
}
```

and then take this configuration object in all methods instead of the separate arguments of abilities, relay_hints and stun_server.

I would keep peer_abilities as a separate argument to transit::init, because that one is determined not via configuration but by examining the peer abilities from the protocol exchange. No need to bother people with it who just want to use `send_file` et al.

Let me know if you're up to it :)

akagifreeez:
Yes, I'm up for it — and I think the split you describe is the right seam. Keeping `peer_abilities` as a separate argument to `transit::init` makes sense too: it's negotiated, not configured.

One thing I'd like to settle before writing it, because it changes user-facing CLI behaviour: **`stun_server: Option<url::Url>`**.

The current value isn't a URL. `DEFAULT_STUN_SERVER` is `"stun.piegames.de:3478"`, and it reaches `transport::tcp_get_external_ip(stun_server: &str)`, which calls `.to_socket_addrs()` on it — so it needs the bare `host:port` form. With `url` 2.x:

```
Url::parse("stun:stun.piegames.de:3478")   -> Ok, host=None, port=None, cannot_be_a_base=true
Url::parse("stun://stun.piegames.de:3478") -> Ok, host=Some("stun.piegames.de"), port=Some(3478)
Url::parse("stun.piegames.de:3478")        -> Ok, scheme="stun.piegames.de", path="3478"
```

The third line is the awkward one: today's `--stun-server stun.piegames.de:3478` would parse *successfully* into something meaningless rather than failing loudly. So adopting `url::Url` means requiring `stun://host:port`, changing the shape of `DEFAULT_STUN_SERVER`, and recombining `host_str()` + `port()` before the `to_socket_addrs()` call.

I'm fine with that — it's consistent with `--relay-server`, which is already a `url::Url` with `tcp://HOSTNAME:PORT`. I just don't want to silently change the flag format if that wasn't the intent. The alternatives would be keeping `Option<String>` inside `TransitConfig`, or a small dedicated type that validates `host:port`.

Whichever you prefer, I'll implement it. I'll also rebase onto main first — the branch is behind since `41d78ae` — and rewrite the CHANGELOG breaking-change entry to describe the config object rather than the individual arguments.

pepa65:
If I may pipe up, I think many users would prefer not to have to use  Option<url::Url> for setting the stun server, and rather stick to the `host:port` format, which is still unambiguous, easier to specify (no need to enter `stun://`) and the current way of doing it. It could easily be a dedicated type to make it clean.

